PR #733, opened 2014-12-02
Integrate new formula into sample size calculator

dtothefp:
@kylerush @darwishoptly This PR addresses adding the new sample size calculator to the marketing website.  As suggested I've integrated a query param `?new=true` that will invoke using the new formula as well as hide inputs that this formula does not utilize. I'm not sure if the output seems correct so I would like it if you guys check out the screenshots and test the page on staging to see if you think the results are correct.

Here is the formula:
https://github.com/optimizely/marketing-website/blob/dfoxpowell/new-sample-size-calculator-formula/website-guts/assets/js/pages/sample-size-calculator.js#L127

(Please not all percents are divided by 100 to normalize.  Ignore it only looks like one is being normalized, the other 2 are being divided by 100 in another function due to dependency on old code);

url: http://optimizely-marketing-site-staging.s3-website-us-east-1.amazonaws.com/dfoxpowell/new-sample-size-calculator-formula/resources/sample-size-calculator/?conversion=3&effect=20&significance=95&new=true (this is the default when loading the page with no query params other than `new=true`)
screenshot:
screen shot 2014-12-02 at 4 32 21 pm

screen shot 2014-12-02 at 4 56 04 pm

url: http://optimizely-marketing-site-staging.s3-website-us-east-1.amazonaws.com/dfoxpowell/new-sample-size-calculator-formula/resources/sample-size-calculator/?conversion=0.001&effect=0.001&significance=95&new=true
screenshot: 
screen shot 2014-12-02 at 4 40 29 pm

screen shot 2014-12-02 at 4 54 57 pm


dtothefp:
@dargani123 @kylerush ###Update please disregard the above comments on testing.  I'm leaving them because in the future if we want to rollback to the old sample size calculator or run an experiment it will be helpful to use the instructions above.  

Anyhow, now to test go to 

http://optimizely-marketing-site-staging.s3-website-us-east-1.amazonaws.com/dfoxpowell/new-sample-size-calculator-formula/resources/sample-size-calculator/

You can choose to enter query parameters to adjust the default values for the calculator 

?conversion=0.1&effect=0.001&significance=85

Or query params will be generated for you automatically as soon as you interact with the calculator.  I tried @dargani123 suggestion to round by 100 by it resulted in bad UX because the user would scroll and sometimes no number change would happen because the 50 interval to round up or down to the nearest 100 hadn't been produced.

Some issues that exist with new content:
- the "Learn More" links for both MDE descriptions link to the same page
- the "Learn More" link for "Where is statistical power in your sample size calculator?" was not specified in the Google Doc https://docs.google.com/a/optimizely.com/document/d/1h98MZ5XJ-KQzS5Ll9r8XrZn5qvl-EyQr-fY1G3th3I4/edit# therefore the current link displays Magnum PI :smile:

PR #751, opened 2014-12-09
Make auto fill smarter.

kylerush:
**Two changes**:

1. Doesn't overwrite values in fields if the fields already have values

2. If `last_name` is falsy, but `first_name` is truthy, the name field will get only `first_name` as the value and vice versa. If both are truthy, the field will get a value of both first name and last name.

**To test**:

Change the `first_name` and `last_name` peroperty values in the `website-guts/endpoint-mocks/accountInfo.json` static JSON file. Restart your server. Verify that the correct values exist on any lead form.

To ensure that the values don't get overwritten, change line 137 of `grunt/connect.js` to: 

```js
setTimeout(function(){
  res.end( grunt.file.read(paths[randIndex]) );
}, 5000);
```

This adds a 5 second delay to the response on the /account/info endpoint. Restart your server. Visit /free-trial and then enter values into the form fields quickly. Verify that the values were not overwritten when the response to /account/info were received.

### DIFF
--- a/website-guts/assets/js/global.js
+++ b/website-guts/assets/js/global.js
@@ -501,31 +501,49 @@ window.optly.mrkt.utils.smoothScroll = function(event) {
 w.optly_q.push([function(){
 	if(typeof w.optly.mrkt.user.acctData === 'object'){
 		if(
+				$('[name="first_name"]').val() === '' &&
 				typeof w.optly.mrkt.user.acctData.first_name === 'string' &&
 				w.optly.mrkt.user.acctData.first_name
+
 			){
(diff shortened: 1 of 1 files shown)


kylerush:
@dtothefp can you review? why you hate my pull requests? :(

dtothefp:
@kylerush I noticed that if the email field is invalid, it still does not get overwritten.  this is probably better the way it is though because otherwise someone could potentially be entering their email and have it be replaced while they are typing.

Also, isn't this a moot point now, because logged in users have no way to actually make it to the ppc page because of that evil redirect?

LGTM :bowtie:

bhamodi:
This is now invalid.

PR #771, opened 2015-01-06
Remove file revving.

kylerush:


tylerbrandt:
assuming blah.cloudfront.net is static looks fine

PR #781, opened 2015-01-13
Stats Engine

ccorda:
Still working, just getting pushed for translation


ccorda:
@kylerush @dtothefp I wasn't quite sure what to do with the video, this seemed like the best way to use their available options and have it tied to the button click.

dtothefp:
@ccorda Just made some slight yml/hbs changes to allow image swapping for translation.  I also will be having @KarenLoughrey add CDN image links into yml so will let you know when that is updated

screen shot 2015-01-19 at 12 01 28 pm

dtothefp:
retest this please

PR #789, opened 2015-01-16
Change image src for segment

stewsmith:
The segment logo on the /partners/technology page is currently broken. This PR fixes that.


dtothefp:
@stewsmith LGTM

PR #796, opened 2015-01-21
MKT-934 -- Send butterbar param to eventbrite when present on /opticon

stewsmith:
https://optimizely.atlassian.net/browse/MKT-934

This PR is so we can see how many people are buying tickets through the butterbar in the product.

It passes the url param aff=butterbar on to eventbrite.

### DIFF
--- a/website-guts/assets/js/pages/opticon.js
+++ b/website-guts/assets/js/pages/opticon.js
@@ -1,3 +1,5 @@
-if (window.optly.mrkt.utils.getURLParameter('aff') === 'butterbar') {
-  $('#register .cta').attr('href', 'https://opticon2015.eventbrite.com/?ref=elink&aff=butterbar');
+var urlParams = window.optly.mrkt.utils.deparam(window.location.search);
+if (urlParams) {
+  var replacementHref = window.optly.mrkt.utils.param('https://opticon2015.eventbrite.com/?ref=elink', urlParams);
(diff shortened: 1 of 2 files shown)


website-guts/assets/js/pages/opticon.js
```diff
@@ -0,0 +1,3 @@
+if (window.optly.mrkt.utils.getURLParameter('aff') === 'butterbar') {
```
jgaul: you could make it more flexible and just say that if any "aff" value is set, that param and its value should be written into the registration href.

stewsmith: Good idea.

dtothefp: @stewsmith you should probably use the deparam function for that then

website-guts/assets/js/pages/opticon.js
```diff
@@ -0,0 +1,5 @@
+var urlParams = window.optly.mrkt.utils.deparam(window.location.search);
+if (urlParams) {
+  var replacementHref = window.optly.mrkt.utils.param('https://opticon2015.eventbrite.com/?ref=elink', urlParams);
```
jgaul: i don't know these utils, but as long as this worked when you tested it, LGTM.

stewsmith: Worked when tested with 0, 1, and multiple params.

PR #797, opened 2015-01-21
Fix stats comparison

nchilada:
I'm not sure why we're making such a misleading comparison.

BTW the footnote doesn't actually show up.  There's just a hidden div with strangely munged text...

### DIFF
--- a/website/statistics/statistics_data.yml
+++ b/website/statistics/statistics_data.yml
@@ -46,8 +46,8 @@ compare:
     id: 2
     title: "With traditional statistics"
     pre: "setting a 95% significance<br>level can result in a"
-    percentage: "30%"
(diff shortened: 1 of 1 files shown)


dtothefp:
LGTM

robinpam:
@dtothefp : Can we please reverse this PR? We want to keep the copy the old way. Thanks.

nchilada:
@robinpam I don't understand how we can justify that comparison.  It's misleading.

robinpam:
Let's talk about it in person. We can't be making changes to the website without running them through marketing first either.

nchilada:
@robinpam There was an email about it, but sure, we should talk in person.

PR #798, opened 2015-01-21
changed sample size formula slightly

darwishoptly:


darwishoptly:
@kylerush @dtothefp going to add my picture to profile now

PR #807, opened 2015-01-23
start of assemble v6 migration

doowb:
- Updating helpers
 - Creating assemble grunt task to wrap assemble v6
 - Updating front-matter to remove `.hbs` for layouts
 - Updating layouts {{> body }} to {% body %}
 - Setting up new assemble tasks with custom renameKey functions for building.

This is just a start and doesn't take advantage of new built-in features.


doowb:
Made migration branch directly.

PR #809, opened 2015-01-23
MKT-825 -- Remove support@ references

stewsmith:
https://optimizely.atlassian.net/browse/MKT-825


dtothefp:
@stewsmith LGTM if the build is fine

PR #819, opened 2015-01-30
Add testing fields for IOS Marketo population test

dtothefp:
From my phone conversation we determined that:

- we could not figure out to change API name within marketo interface for custom fields
- we are unsure if changing field name in SalesForce resulted in changing the Munchkin API name for the custom field
- we created 2 new custom fields to test the IOS checkbox
- 1 with default created name by Marketo, this seems to have no underscores and lowercase first letter
- 2 we altered the default name and Capitalized the first letter.

#### Default API Name
screen shot 2015-01-30 at 2 05 52 pm

#### Custom Capitalized API Name
screen shot 2015-01-30 at 2 17 29 pm


kylerush:
LGTM

PR #820, opened 2015-01-30
Move registration section towards the top

stewsmith:
We want the registration section towards the top so this PR does that.

### DIFF
--- a/website/opticon/index.hbs
+++ b/website/opticon/index.hbs
@@ -40,35 +40,6 @@ external_stylesheets:
     </div>
   </section>
 
-
-  <!--                         -->
-  <!--Call for Speakers Section-->
-  <!--                         -->
-  <section id="call-for-speakers" class="call-for-speakers-panel">
-    <h1 class="section-title">{{call_for_speakers.title}}</h1>
-    <p class="blurb">{{{call_for_speakers.blurb}}}</p>
-  {{/with}}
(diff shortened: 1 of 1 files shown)


website/opticon/index.hbs
```diff
@@ -40,35 +40,6 @@ external_stylesheets:
...
-    <h1 class="section-title">{{call_for_speakers.title}}</h1>
-    <p class="blurb">{{{call_for_speakers.blurb}}}</p>
-  {{/with}}
-  <div class="button-container desktop">
-    <a class="cta" href="{{linkPath}}/opticon/speakers">{{opticon_data.call_for_speakers.cta.desktop}}</a>
-  </div>
-  <div class="button-container mobile">
-    <a class="cta" href="{{linkPath}}/opticon/speakers">{{opticon_data.call_for_speakers.cta.mobile}}</a>
-  </div>
-  {{#with opticon_data}}
```
chestone: What happened to this? Was this data modified? Looks like we are pulling from call_for_speakers as a global object rather than opticon_data.call_for_speakers.

chestone:
Outside of that comment it looks good to me!

website/opticon/index.hbs
stewsmith: Good catch. I removed a redundant {{with}}

PR #821, opened 2015-01-30
Add more tracking events for an experiment.

kylerush:
This PR is a preparation for an experiment that we are going to run on the mobile product page for ads audience. We are adding more reporting.

Reporting changes to the /mobile page:
1. Added blur reporting
2. Added focus reporting

Reporting changes to the /free-trial page:
1. Changed "blur" to "focus to match the real event happening
2. Added blur reporting

Note that for the /mobile page, since there are two forms, I'm sending out two events, one specific to the form on which it occurred and another that is not specific the form on which it occurred. The latter event helps us understand aggregate behavior and is particularly necessary for the upcoming experiment.

### DIFF
--- a/website-guts/assets/js/layouts/trial-form.js
+++ b/website-guts/assets/js/layouts/trial-form.js
@@ -12,9 +12,25 @@ var xhrInitiationTime;
 
 //track focus on form fields
 $('#seo-form input:not([type="hidden"])').each(function(){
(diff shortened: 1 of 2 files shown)


kylerush:
@dtothefp can you review?

kylerush:
@stewsmith can you review?

stewsmith:
LGTM

PR #822, opened 2015-02-02
Update copy on website for Android beta launch.

kylerush:


kylerush:
@seanoliver this will be live in prod momentarily

PR #823, opened 2015-02-03
add moment helper to fix copyright year in footer

dtothefp:
Addresses https://optimizely.atlassian.net/browse/MKT-1006 to dynamically update to the current year for copyright in the footer.

*Note: that upgrading to Assemble v6 we will have to upgrade this helpers https://github.com/helpers/helper-moment/issues/1*


stewsmith:
LGTM

PR #827, opened 2015-02-05
Adds website packaging support to release assets into optimizely/ via grunt release

brianlucas:
This allows you to bundle up your website release and push it out to github.  You can check the releases here:

https://github.com/optimizely/marketing-website/releases

To run this process, type 'grunt release' and enter your username and API password.  You may need to create a github API password here:
https://github.com/settings/applications


brianlucas:
cc @kylerush

kylerush:
@brianlucas i'm not sure it's practical for us to run this for every merge we do, is that the expectation? i've used tags before which work great for bigger applications with fewer deploys, but we deploy often and we deploy pull requests from marketers that don't work in the command line and we don't test their stuff in the command line. so this would add quite a bit of bulk to our process. do i have that wrong?

brianlucas:
No, it’s completely optional and can be run anytime you want to push updates into the optimizely app (for testing, etc).

It doesn’t need to be run with every merge or release.

kylerush:
@brianlucas gotcha, thanks

package.json
```diff
@@ -47,9 +47,13 @@
   },
   "devDependencies": {
     "body-parser": "1.8.3",
+    "grunt-contrib-compress": "^0.13.0",
```
dtothefp: @brianlucas we like to constrain all our dependency versions, can you remove the carets?

Gruntfile.js
```diff
@@ -133,5 +135,23 @@ module.exports = function(grunt) {
   grunt.registerTask('default', [
     'build'
   ]);
```
dtothefp: @brianlucas I've never used `grunt.loadNpmTasks`, what does it do?

dtothefp:
@brianlucas LGTM as soon as you constrain the package versions in package.json :neckbeard:

PR #832, opened 2015-02-07
Adjust redirect on om pages.

kylerush:
Per MKT-1126


kylerush:
@dtothefp can you review?

PR #833, opened 2015-02-09
Enhance error tracking for mobile experiment.

kylerush:
This gives us the ability to group error on both the top and bottom forms. Without this we couldn't track overall error rate in Optimizely during our next experiment. In the past, on validation error, only this was sent: `mobile-signup-form-top password1 error submit`. With this PR, in addition to that event, it will also send `mobile-signup-form password1 error submit`.

It also ads a submit tracking event in the before function. This will help us measure the `submit : success` ratio.

### DIFF
--- a/website-guts/assets/js/pages/mobile.js
+++ b/website-guts/assets/js/pages/mobile.js
@@ -206,6 +206,14 @@ $(function() {
     signupMobileMvppTopHelperInst.processingAdd();
     signupMobileMvppTopHelperInst.removeErrors();
     signupMobileMvppTopHelperInst.optionsErrorElm.innerHTML = signupMobileMvppTopHelperInst.errorMessages.DEFAULT;
+    w.analytics.track('/mobile/submit', {
+      category: 'account',
+      label: w.optly.mrkt.utils.trimTrailingSlash(w.location.pathname)
+    }, {
+      integrations: {
+        'Marketo': false
+      }
+    });
     return true;
   });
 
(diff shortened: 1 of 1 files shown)


kylerush:
@david-optimizely can you review?

website-guts/assets/js/pages/mobile.js
```diff
@@ -175,6 +175,18 @@ $(function() {
...
 
+  //custom validation error reporting handler
+  var validationErrorCustom = function(element){
+    var elementName = element.getAttribute('name');
+    var elementValue = $(element).val();
+    var elementHasValue = elementValue ? 'has value' : 'no value';
+    w.analytics.track('mobile-signup-form ' + elementName + ' error submit', {
+      category: 'form error',
+      label: elementHasValue,
+      value: elementValue.length
```
dtothefp: @kylerush is there a reason that the value is the length, rather than the actual element value

website-guts/assets/js/pages/mobile.js
```diff
@@ -194,10 +206,19 @@ $(function() {
...
+        'Marketo': false
+      }
+    });
     return true;
   });
 
-  signupFormTop.on('validationerror', function(elm) {
+  signupFormTop.on('validationerror', function(elm){
+    validationErrorCustom(elm);
     w.optly.mrkt.Oform.validationError(elm);
```
dtothefp: @kylerush isn't `w.optly.mrkt.Oform.validationError(elm);` doing the exact same thing as `validationErrorCustom(elm);` only the new function doesn't report the form id and is specific to mobile signup form? Assuming that is the intention from your PR description

website-guts/assets/js/pages/mobile.js
kylerush: @dtothefp yes, the value attribute is a number, not a string. given that restriction i'm using the attribute to record how many characters people enter into the fields to give me slightly more information before the errors.

dtothefp:
@kylerush left a couple notes. Don't think there is anything to fix just had some questions. Otherwise, LGTM

website-guts/assets/js/pages/mobile.js
kylerush: @dtothefp yes, this is intentional.

PR #844, opened 2015-02-10
Adjust PPC redirect.

kylerush:
Per MKT-1126

To test:

1. Open your host file
2. Point both wwww.optimizely.com and www.optimizely.de to your local server. Example: 
```
0.0.0.0 www.optimizely.com
0.0.0.0 www.optimizely.de
```
3. Save the changes to your host file
4. Visit both www.optimizely.de:9000/free-trial and www.optimizely.com:9000/free-trial
5. Submit the form. You should be taken to the german editor on .de when you submit on the German page and the US editor on .com when you submit on the .com domain.
6. MAKE SURE YOU REVERT THE CHANGES TO YOUR HOST FILE

### DIFF
--- a/website-guts/assets/js/om/layouts/original.js
+++ b/website-guts/assets/js/om/layouts/original.js
@@ -117,8 +117,10 @@ w.optly.mrkt.trialForm = new Oform({
 
       if(!w.optly.mrkt.automatedTest()){
         setTimeout(function(){
-          var redirectURL, domain;
+          var redirectURL, domain, queryParams;
           domain = window.location.hostname;
+          queryParams = window.location.href.split('?', 1)[1] || '';
+          queryParams = queryParams ? '&' + queryParams : queryParams;
           if(/^www\.optimizely\./.test(domain)){
             //production
             redirectURL = '/edit?url=';
@@ -126,7 +128,7 @@ w.optly.mrkt.trialForm = new Oform({
             //local dev
             redirectURL = 'https://www.optimizely.com/edit?url=';
           }
-          w.location = redirectURL + encodeURIComponent(d.getElementById('url').value) + '&' + window.location.href.split('?', 1)[1];
+          w.location = redirectURL + encodeURIComponent(d.getElementById('url').value) + queryParams;
         }, 1000);
       }
 
--- a/website-guts/assets/js/om/layouts/original.js
+++ b/website-guts/assets/js/om/layouts/original.js
@@ -119,7 +119,7 @@ w.optly.mrkt.trialForm = new Oform({
         setTimeout(function(){
           var redirectURL, domain, queryParams;
           domain = window.location.hostname;
-          queryParams = window.location.href.split('?', 1)[1] || '';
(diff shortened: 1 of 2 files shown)


kylerush:
@stewsmith can you review?

website-guts/assets/js/om/layouts/original.js
```diff
@@ -117,7 +117,13 @@ w.optly.mrkt.trialForm = new Oform({
...
       if(!w.optly.mrkt.automatedTest()){
         setTimeout(function(){
-          w.location = 'https://www.optimizely.com/edit?url=' + encodeURIComponent(d.getElementById('url').value) + '&' + window.location.href.split('?')[1];
+          var redirectURL;
+          if(/^www\.optimizely\./.test(window.location.hostname)){
+            redirectURL = '/';
+          } else {
+            redirectURL = 'https://www.optimizely.com/edit?url=';
+          }
+          w.location = redirectURL + encodeURIComponent(d.getElementById('url').value) + '&' + window.location.href.split('?')[1];
```
stewsmith: The `window.optly.mrkt.utils.deparam` function gets all the parameters instead of doing `window.location.href.split('?')[1]`.

Also, why do you use `window.location.href.split('?')[1]` only in the om version and not trial-form.js?

kylerush: @stewsmith doesn't deparam turn the url params into an object? what i'm doing here gets all the params as a string. seems like deparam would be a more work since i'd have to run deparam to turn the query strings into an object from a string and then turn them back into a string. seems like the way i'm doing it now is more efficient. i think deparam is more if you want to read the values of the url params, which i am not doing here. i am just copying them.

i'm only using `window.location.href.split` on the new one because for some reason it's the only ay i can get attribution to work. i've been trying to fix that for a week and this is the only solution that i've found. eventually once segment fixes their marketo integration we can remove this.

stewsmith:
LGTM once the OM version is tested.

website-guts/assets/js/om/layouts/original.js
```diff
@@ -117,7 +117,18 @@ w.optly.mrkt.trialForm = new Oform({
 
       if(!w.optly.mrkt.automatedTest()){
         setTimeout(function(){
-          w.location = 'https://www.optimizely.com/edit?url=' + encodeURIComponent(d.getElementById('url').value) + '&' + window.location.href.split('?')[1];
+          var redirectURL, domain, queryParams;
+          domain = window.location.hostname;
+          queryParams = window.location.href.split(/\?(.+)?/)[1] || '';
```
stewsmith: Why do we only want the first param here?

kylerush: @stewsmith that regex splits the string on the first "?" and turns the two strings into an array. so the [0] key is everything before the "?" and the [1] key is everything after.

stewsmith: :+1:

PR #847, opened 2015-02-11
MKT-822 --Remove ad tracking from wrapper

stewsmith:
Per [MKT-822](https://optimizely.atlassian.net/browse/MKT-822)


kylerush:
@stewsmith we should be leaving linked in and doubleclick, but taking out the last tag (adwords/google mobile)

PR #848, opened 2015-02-11
MKT-822 -- remove only google adwords script

stewsmith:
[MKT-822](https://optimizely.atlassian.net/browse/MKT-822)


kylerush:
LGTM

PR #852, opened 2015-02-12
Fix some things.

kylerush:
Changes:
 - Adds a global variable for scroll tracking goal for the experiment
- Cleans up some spacing on type
- Use `&raquo;` instead of `>>`
- Added "Your current plan"
- Prevent downgrade from enterprise

### DIFF
--- a/website-guts/endpoint-mocks/accountInfo.json
+++ b/website-guts/endpoint-mocks/accountInfo.json
@@ -14,7 +14,7 @@
     "trusted_tester": false,
     "beta_customer": false,
     "has_seen_first_notice": false,
-    "plan_id": "enterprise",
+    "plan_id": "free_light",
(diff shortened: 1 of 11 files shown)


kylerush:
@stewsmith can you review?

website-guts/endpoint-mocks/accountInfo.json
```diff
@@ -14,7 +14,7 @@
     "trusted_tester": false,
     "beta_customer": false,
     "has_seen_first_notice": false,
-    "plan_id": null,
+    "plan_id": "free_light",
```
stewsmith: Is this intended to be permanent?

stewsmith:
@kylerush LGTM after you confirm `plan_id` should be kept as `free_light`

PR #853, opened 2015-02-12
Change down arrow from svg to png

stewsmith:


stewsmith:
@kylerush Can you review?

I've tested in Chrome, FF, Safari, IE10 &IE11

PR #854, opened 2015-02-12
Move pricing2 to pricing

stewsmith:
Merge this PR when we want to move /pricing2 to /pricing

@kylerush The experiment needs to begin shortly after.


kylerush:
LGTM

PR #858, opened 2015-02-17
Fade-in for CTA on homepage.

kylerush:
This PR adds a small delay and fade-in for the CTA on the homepage. This is so that we can pull the Demandbase information and update the CTA language before the content fades in. It helps us avoid the so-called flash of unstyled content.


kylerush:
@dtothefp can you review por favor?

dtothefp:
@kylerush nice animation speed per EH :neckbeard: 

LGTM

PR #859, opened 2015-02-18
Modify OM reporting

kylerush:
Three changes:

1. Doesn't show validation error on blur if the field has no value (this is how the old page does it). To test focus into both the email and URL fields, don't enter a value, then focus out, you should not see any red/validation error. Then enter an invalid value in URL and email fields (like "t"), focus out, then see the validation errors. 

2. Adds a `seo-form success after error true` or `seo-form success after error false` to track if the person that submitted the form successfully encountered a validation error. This can only be tested in production.

3. Fixes/adds focus and blur reporting. You can see in the diff that the existing focus reporting was fucked up because it was set to track blur, but was reporting focus.

4. Adds `change engagement` events to track when the user puts a value into each and any form field.

### DIFF
--- a/website-guts/assets/js/om/layouts/original.js
+++ b/website-guts/assets/js/om/layouts/original.js
@@ -10,6 +10,8 @@ d.getElementById('hidden').value = 'touched';
 
 var xhrInitiationTime;
 
+w.optly.mrkt.formHadError = false;
+
 //track focus on form fields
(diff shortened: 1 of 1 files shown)


dtothefp:
*unrelated* logo is breaking in tablet/mobile
screen shot 2015-02-18 at 1 05 51 pm

PR #863, opened 2015-02-19
Add user timing for Optimizely snippet.

kylerush:
With this change we can start monitoring the amount of time it takes the Optimizely snippet to load using Google Analytics user timings.


kylerush:
@dtothefp can you review?

dtothefp:
@kylerush do you want this in the OM wrapper and globals as well? otherwise :bowtie: LGTM

dtothefp:
@kylerush LGTM :bowtie:

PR #873, opened 2015-02-25
Opticon V2 Site

stewsmith:


stewsmith:
@dtothefp Could you look at the following:
1) Font application -- I'm almost certain there's a better way to do this. I'm putting the same font on way too many classes.
2) The anim-opticon-button() mixin. It's a bit jank and works inconsistently on the homepage.
3) The confetti images. I've got them positioned absolute and one has a top: 46em. Maybe there's a better way?
4) The title on the homepage isn't flush ([image](http://cl.ly/image/3I0M2t1w0A0V)). It might be helpful to look at this in person.

Could you ignore the following:
How fonts display -- still need to bring in the external stylesheet.

Thanks!

dtothefp:
weird side scroll issue on about page http://cl.ly/image/270n2R0F2L2I

These buttons are cropped bad, i.e. flat on top and bottom, are we getting new ones http://cl.ly/image/0w3X3O1k2v2S

Padding issues for tables in tablet size on /agenda page http://cl.ly/image/2n2Z1k2N2i3G, also centering issues on same page in mobile.

Spacing a bit "wonk"  on /agenda in mobile for top nav http://cl.ly/image/2m102L1x2K0m

On the grids on /about in my opinion it is generally always a better idea to have `width: 50%` and then pad the inside for your spacing. This is less brittle and won't break with differences in browsers. Also, are we thinking of having `float` fallbacks from now on for older browsers?  http://cl.ly/image/2S0J2a3K3A2a

Same thing with the grids here http://cl.ly/image/0P2Q3M211T1g. I see they also both have `flex: 1` which I think you would really only want to use if one grid item is constrained in width and the other you want to grow and shrink to fit the remaining space.

For the "confetti image" issue I would think about setting up your page in sections.  For /about you have 
- hero
- top grid
- reg button blue
 - FAQ grid
- bottom grey boxes
This would allow you to keep consistent padding on sections as we talked about that before. Looks like you did that towards the bottom half but not the top. If you did if for the top then you can `position: relative` the hero section, and then `position: absolute` the confetti inside of that. I would generally shy away from `position: absolute` without a bounding container.

dtothefp:
@stewsmith link to `optimizely` on the homepage is messed up because it is relative 

screen shot 2015-03-09 at 4 57 03 pm

website-guts/assets/css/layouts/opticon.scss
```diff
@@ -0,0 +1,215 @@
...
+    .cta-arrow {
+      right: 0px;
+    }
+    .cta-text {
+      left: 20px;
+    }
+    background-color: #E9F3F4;
+  }
+}
+
```
dtothefp: @stewsmith why are these still mixins? Not launch blocking, just would probably be more semantically correct as classes that are extended.

.nvm
```diff
@@ -0,0 +1 @@
+Subproject commit 64de890d079e98bc62ba16c7d906dcf06ea84cc3
```
dtothefp: @stewsmith did you figure out what this is?

website-guts/assets/css/layouts/opticon.scss
```diff
@@ -0,0 +1,215 @@
...
+  .confetti-left {
+    @include opticon-confetti('//d1qmdf3vop2l07.cloudfront.net/optimizely-marketer-assets.cloudvent.net/raw/opticon/pattern_left.svg', 450px, 450px);
+    display: none;
+  }
+  .confetti-right {
+    @include opticon-confetti('//d1qmdf3vop2l07.cloudfront.net/optimizely-marketer-assets.cloudvent.net/raw/opticon/pattern_right.svg', 227px, 400px);
+    right: 0;
+    display: none;
+  }
+}
```
dtothefp: @stewsmith this is definitely more appropriate use of a mixin, good idea

website-guts/assets/css/modules/opticon-footer.scss
```diff
@@ -0,0 +1,61 @@
...
+    &.body-button {
+      color: $opticon-brand-blue;
+    }
+  }
+  .social {
+    background-repeat: no-repeat;
+    background-size: contain;
+    width: 35px;
+    padding-top: 35px;
+    margin: 10px;
```
dtothefp: @stewsmith you really only need to use `padding-top` if you want it responsive. with fixed width you can just use height, but it doesn't really matter either way

dtothefp:
@stewsmith looks pretty good overall...css is starting to get a lot more sexy :+1:

dtothefp:
LGTM

PR #883, opened 2015-02-28
add dest checking helper and smooth scroll for opticon nav

dtothefp:
@stewsmith I made a mistake, in the layout you have to use `page.dest` and in there its a string not an object like in a page. 

#### To Test:
- add another directory called `opticon-2` or whatever.
- copy the opticon page in there
- start the server and go to `/opticon`. Smooth scroll should work
- go to `/opticon-2` and links should redirect to opticon at the selected links place on the page.


stewsmith:
LGTM

PR #884, opened 2015-03-02
Fix error reporting in concat banner and account for root directory in trim utility function

dtothefp:
This PR addresses inproper variable naming for error reporting in our concat banner try/catch blocks. The error was also present in the the `trackLead` response parsing. I noticed this by improper reporting in GA in the screenshot below:

#### To Test
- to test locally you would have to copy and past the concat try/catch block into the `dev` concat banner and then intentionally throw an error in the JS bundle and set breakpoints in the inspector to ensure proper error parsing.
- to test the trimpath function got the console and `window.optly.mrkt.utils.trimTrailingSlash('/')`

screen shot 2015-03-02 at 12 02 36 pm


kylerush:
LGTM

PR #890, opened 2015-03-05
Seeing Red

stewsmith:
This PR removes the anonymous wall js, the anonymous wall index, and the old homepage js.


kylerush:
LGTM

PR #899, opened 2015-03-06
remove munchkin from track lead

dtothefp:
This PR removes the `munchkin` calls we were making to `associateLead` because this is now being taken care of by segment.

In addition if fixes two errors on the `om` for the `err` variable and the `XHR` property on the `XHRevent` object

### DIFF
--- a/website-guts/assets/js/om/utils/oform-globals.js
+++ b/website-guts/assets/js/om/utils/oform-globals.js
@@ -50,11 +50,11 @@ w.optly.mrkt.Oform.trackLead = function(args){
   source = w.optly.mrkt.source;
 
   try {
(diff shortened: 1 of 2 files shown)


website-guts/assets/js/om/utils/oform-globals.js
```diff
@@ -50,11 +50,11 @@ w.optly.mrkt.Oform.trackLead = function(args){
   source = w.optly.mrkt.source;
 
   try {
-    response = JSON.parse(XHRevent.XHR.responseText);
+    response = JSON.parse(XHRevent.responseText);
```
kylerush: @dtothefp this should be `XHRevent.XHR.responseTest`

kylerush:
LGTM

PR #900, opened 2015-03-07
Add om-tests to grunt test and grunt ui-test.

kylerush:
Last Thursday I deployed a bug to production that broke all leads on the /om/free-trial page. The new OM tests would have caught this, but for some reason neither David nor I ran the OM tests on the branch so we didn't catch the bug. Obviously we can't afford anymore lead bugs so we need to ensure that the OM tests are always run on every PR.

To prevent this from happening in the future, this PR integrates the OM tests into the `grunt test` and `grunt ui-test` tasks. That means you can run the OM tests in three different ways:

1. `grunt test` while your local sever is NOT running
2. 'grunt ui-test` while your local server IS running - this run all UI tests
3. `grunt ui-test:om` while your local server IS running - this runs only the OM tests

There isn't much of a change for anyone here. As mentioned above, the OM tests will now run if you were running the UI tests with either `grunt test` or `grunt ui-test`. The only change is that you'll now have to verify that the OM tests pass.

**Make sure you run `npm i` after this is merged.**

This is what a passing OM test looks like (no red):

screen shot 2015-03-07 at 10 09 13 am

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -21,7 +21,8 @@ module.exports = function(grunt) {
         resemble: 'grunt-resemble-cli',
         sass: 'grunt-sass',
         connect: 'grunt-contrib-connect',
-        jasmine_node: 'grunt-jasmine-node'
+        jasmine_node: 'grunt-jasmine-node',
+        open: 'grunt-open'
       }
     },
     data: {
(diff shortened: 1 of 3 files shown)


kylerush:
@dtothefp @stewsmith moar tests. read the description por favor.

dtothefp:
LGTM

PR #904, opened 2015-03-09
Change priority so stars are in order

stewsmith:
Partner stars were out of order. This PR puts partners in order by star values.


dtothefp:
@stewsmith LGTM

PR #908, opened 2015-03-11
Add prospectus to about page

stewsmith:
We need a link to the sponsorship prospectus. This PR replaces the Opties paragraph (which lacks a CTA) with the prospectus.


danoc:
Fixed heights are nasty, but LGTM otherwise! :+1:

PR #920, opened 2015-03-16
remove multichannel experiences from mobile LP

seanoliver:
this feature is not supported by the product


dtothefp:
@seanoliver looks good

PR #924, opened 2015-03-17
add urlLike helper to check for valid KB article url

dtothefp:
###Purpose
- not allow marketers to enter invalid input for `kb_article` parameter. This is the link to the external article button.
- if the url is not valid https, http, or //something.co the button will not be displayed

###Test
Please test this, the regex was breaking Node build previously
- just run `grunt server as normal`
- got to `partners/technology/liveramp/` and check there is no "Get implementation guide" button
- got to another technology partner page and ensure the button is there with functioning link

screen shot 2015-03-17 at 12 49 32 pm

### DIFF
--- a/website-guts/helpers/helper-debug.js
+++ b/website-guts/helpers/helper-debug.js
@@ -1,0 +1,13 @@
+module.exports.register = function (Handlebars)  {
+  Handlebars.registerHelper('debug', function (optionalValue, options)  {
+    console.log('Current Context');
+    console.log('====================');
+    // console.log(this);
+    if (optionalValue) {
+      console.log('Value');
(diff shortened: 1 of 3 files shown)


website-guts/helpers/helper-debug.js
```diff
@@ -0,0 +1,13 @@
+module.exports.register = function (Handlebars)  {
+  Handlebars.registerHelper('debug', function (optionalValue, options)  {
+    console.log('Current Context');
+    console.log('====================');
+    // console.log(this);
```
stewsmith: uncomment this?

stewsmith:
LGTM

PR #929, opened 2015-03-19
MKT-1572 -- Add api terms to new page

stewsmith:
New page for legal API terms

[Ticket](https://optimizely.atlassian.net/browse/MKT-1572)


dtothefp:
@stewsmith could you do this in external YML file, I know it sucks but it's better for translation. You can use `{{isnt @first}}` I think to add the padding and h4

dtothefp:
@stewsmith LGTM

PR #931, opened 2015-03-19
Move api-terms to terms-api

stewsmith:


dtothefp:
@stewsmith LGTM after it builds

PR #938, opened 2015-03-24
Hound config

stewsmith:
# The following scss linting rules will now be enforced via hound:

For a more complete listing see: 
[Linting Options](https://github.com/causes/scss-lint/blob/master/lib/scss_lint/linter/README.md#urlquotes)

## Bang Format
Reports when you use improper spacing around ! (the "bang") in !important and !default declarations.

## Color Keyword
Prefer hexadecimal color codes over color keywords.

## Color Variable
Prefer color literals (keywords or hexadecimal codes) to be used only in variable declarations. They should be referred to via variables everywhere else.

## Comment
Prefer `//` comments over `/* ... */`.

## Debug Statement
Reports @debug statements

## Duplicate Property
Reports when you define the same property twice in a single rule set.

## Empty Rule
Reports when you have an empty rule set.

## Hex Notation
Checks if hexadecimal colors are written in lowercase.

## Hex Validation
Ensure hexadecimal colors are valid

## Indentation
Use two spaces per indentation level.

## Mergeable Selector
Reports when you define the same selector twice in a single sheet.

## Property Count
Limit the number of properties in a rule set.

## Property Spelling
Reports when you use an unknown CSS property

## Qualifying Element
Avoid qualifying elements in selectors
Bad:
```
div#thing {
  ...
}

ul.list {
  ...
}
```

## Shorthand
Prefer the shortest shorthand form possible for properties that support it.

## Single line per property
Properties within rule sets should each reside on their own line.

## Single line per Selector SpaceAfterComma SpaceAfterPropertyColon SpaceBetweenParens SpaceBeforeBrace SpaceAfterPropertyName
Conventional spacing

## UrlQuotes
URLs should always be enclosed within quotes.

##   VendorPrefixes:
Avoid vendor prefixes. That is, don't write them yourself.

##   ZeroUnit
Omit length units on zero values.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -3,6 +3,7 @@ java_script:
 
 scss:
   enabled: true
+  config_file: .scss-style.yml
 
 ruby:
   enabled: false
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,12 @@
+java_script:
+  enabled: false
+
+scss:
+  enabled: true
+
+ruby:
+  enabled: false
+
+coffee_script:
+  enabled: false
+
--- a/.scss-style.yml
+++ b/.scss-style.yml
@@ -1,0 +1,145 @@
+scss_files: "**/*.scss"
+linters:
+  BangFormat:
+    enabled: true
+    space_before_bang: true
+    space_after_bang: false
+  BorderZero:
+    enabled: false
+    convention: zero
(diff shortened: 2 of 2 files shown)


dtothefp:
does this mean that we are preferring hex over rgb Color Variable?

dtothefp:
LGTM

PR #945, opened 2015-03-25
Opticon Agenda

stewsmith:
# Opticon Agenda
https://optimizely.atlassian.net/browse/CD-757

The agenda has the following functionality:
1. Filter talks by Track and Roles
2. Expand and collapse talks accordion style
3. Navigate with Day 1 and Day 2 links
4. Filter and Day links become fixed to top upon scroll

## Things to note:
1. <strong> CSS </strong>
  1. The filter pill reuses the pill from [Solution Partners](https://www.optimizely.com/partners/solutions/). 
  2. There is a rotation transform on the ❯ upon expanding a talk
2. <strong>  JS </strong> 
  1. The accordion for the talks is pretty straightforward.
  2. `w.optly.mrkt.utils.smoothScroll` is overridden. There is probably a better way to do this.
  3. The scroll logic for highlighting the Days at the top is a bit complicated. There is a good chance I could do something better here.
  4. The filter logic begins on line 135, and uses `changeTalkVisibility`, `isSubset`, and `toggleArrayItem`
3. <strong> YML</strong>
  1. This file is crazy long. The content is ~~filler content~~ updated. We are not launching with description or location. These will be added once they are known.


website-guts/assets/css/pages/opticon-agenda.scss
```diff
@@ -1,70 +1,143 @@
...
-    padding: 5px 20px 0 5px;
-    font-size: 0.875em;
-    color: $almost-black-text-color;
-    letter-spacing: 1px;
-    span {
-      font-size: 0.6875em;
+  .up-arrow {
+    &:after { content: '   ↑'; }
+  }
+  .pill--opticon {
```
dtothefp: @stewsmith I think rather than having whitespace in the pseudo element content you may want to just position it `relative` where you want it

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
...
+function calculateDayTwoDistanceFromTop() {
+  dayTwoDistanceFromTop = $dayTwo.offset().top - $window.scrollTop();
+}
+
+//Accordian for talks
+$('.js-toggle-cont').slideToggle('fast'); //all details closed to start
+$('.js-accordian-control').click(function(event) {
+  $(this).nextAll('.js-arrow').toggleClass('expansion-arrow--open');
+  $(this).nextAll('.js-toggle-cont').slideToggle('fast', calculateDayTwoDistanceFromTop);
+});
```
dtothefp: @stewsmith instead of `nextAll` you could use `.find` because you are only looking for one thing and not many

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
+var dayTwoDistanceFromTop = 3319,
+    $window = $(window),
+    $dayTwo = $('#day-two');
+/*
```
dtothefp: @stewsmith seems weird to be setting a fixed distance for `dayTwoDistanceFromTop` shouldn't you just measure it with jQuery `offset().top`?

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
...
+      $dayTwoNav.removeClass('agenda-day--active up-arrow');
+    } else if (dayTwoDistanceFromTop < 150) { // activate day two
+      $dayTwoNav.addClass('agenda-day--active');
+      $dayOneNav.removeClass('agenda-day--active up-arrow');
+      if (dayTwoDistanceFromTop < 25) {
+        $dayTwoNav.addClass('up-arrow');
+      }
+    }
+  }
+});
```
dtothefp: @stewsmith seems redundant to be calling all these add and remove classes over and over when they only need to be applied once when the state of less than or more than 200 changes. I'm sure you could create some sort of variable for caching state and do something like `if (scrollTop < 200 && cachedStateVar)`

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
...
+
+/*
+  If item is in array, remove it
+  else, add it
+*/
+function toggleArrayItem(array, item) {
+  var itemIndex = $.inArray(item, array);
+  if(itemIndex !== -1) {
```
dtothefp: @stewsmith why not just use raw JS here `array.indexOf(item)`?

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
...
+  Check that every item in arr1
+  is in arr2 i.e. arr1 is subset
+  of arr2
+*/
+function isSubset(arr1, arr2) {
+  return arr1.every(function(val) {
+    return arr2.indexOf(val) >= 0;
+  });
+}
+
```
dtothefp: @stewsmith this is cool, I didn't know the `every` method

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
...
+//Top filter dropdown
+$dropdownElems.click(function(event) {
+  var $this = $(this);
+  $this.toggleClass('active');
+  $dropdownElems.not( $this ).removeClass( 'active' );
+  $this.mouseleave(function(){
+    $this.removeClass( 'active' );
+  });
+});
+
```
dtothefp: @stewsmith maybe do something like this because seems like you are setting the event over and over

```js
$this.one('mouseleave', function(){
   $this.removeClass( 'active' );
 });
```

or just register the mouseleave event outside of the click event so it is only getting registered once.

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    // If no talks match filter criteria,
+    // hide the time slot
+    if (visibleTalks.length === 0) {
+      $(this).parent().hide();
+    } else {
+      $(this).parent().show();
+    }
+
```
dtothefp: @stewsmith I think it would be nice to hide/show with an animation

could either use jquery http://api.jquery.com/hide/ or just animate the height of the element with CSS. I have an animation library and SASS mixin for this we could use if you want to try https://github.com/optimizely/marketing-website/wiki/Global-Animation-Service

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
...
+    visibleTalks.each(function(index, event) {
+      var $event = $(event);
+      if (index === 0) {
+        $event.css('margin-top', 0);
+      } else {
+        $event.css('margin-top', '50px');
+      }
```
dtothefp: @stewsmith you could do this with raw JS `event.style.marginTop = 0;`. Also, maybe think about renaming the variable `event` because it's kind of confusing because as it's generally used with a function that passes on an event

website-guts/assets/js/pages/opticon-agenda.js
```diff
@@ -0,0 +1,168 @@
...
+    $events.each(function(eventIndex, eventItem) {
+      // array of tracks and roles for each event
+      var eventFilterList = ($(eventItem).data('track') + ' ' + $(eventItem).data('roles')).split(' ');
+      // if every item in filterlist isn't present on the event, hide it
+      if(!isSubset(filterList, eventFilterList)) {
+        $(eventItem).hide();
+      } else {
+        $(eventItem).show();
+      }
+    });
```
dtothefp: @stewsmith again maybe we do an animation here

dtothefp:
@stewsmith looks good. We should talk about whether we should animate the items when they are shown/hidden

website-guts/assets/js/pages/opticon-agenda.js
stewsmith: `.find` will search the descendants whereas I want the sibling. `next` also doesn't work in this case.

website-guts/assets/css/pages/opticon-agenda.scss
stewsmith: Good idea

website-guts/assets/js/pages/opticon-agenda.js
stewsmith: Yes good idea.

website-guts/assets/js/pages/opticon-agenda.js
stewsmith: Good call.

dtothefp:
@stewsmith looks good for the animation. There's lots of other leftover comments you could clean up but it's not really launch blocking

PR #961, opened 2015-03-31
Add signed in custom dimension.

kylerush:
Adds a "Signed in" custom dimension with values of "true" or "false". This will help us determine how many visits to the marketing website are from signed in visitors. I added a test for this. All tests pass with `grunt test`.


kylerush:
@stewsmith can you review?

stewsmith:
LGTM

PR #973, opened 2015-04-08
fix selector for leadform type and add fallback

dtothefp:
on all benefits pages
- /demo
- /enterprises
- /ecommerce

the `initContactForm` method was referencing a selector for `InboundLeadformType` that didn't exist

#### to test

submit form on any of these pages and put a breakpoint in the `w.optly.mrkt.utils.oform.initContactForm` method and check the object that is being passed to the `identify` method


stewsmith:
LGTM

PR #971, opened 2015-04-07
Create index.hbs

Hiliaryrobertson:
@julieritchie @kellyfinn Please review.

PR #974, opened 2015-04-08
Fresh Egg AU Solution Partner

dtothefp:
@Hiliaryrobertson @stewsmith @kellyfinn I migrated https://github.com/optimizely/marketing-website/pull/971 to here because it was being merged into the wrong location


dtothefp:
@kellyfinn can you please add the logo? @stewsmith I created this as a new branch and updated to new Assemble syntax

stewsmith:
@kellyfinn Could you take a look at this PR? Thanks!

kellyfinn:
@dtothefp @stewsmith 
I'm not sure why the logo isn't appearing for me in the preview. Thanks for taking a look. Also FYI we will be merging listings such as these that are listed multiple times for various locations into one single listing.  For example, Fresh Egg has a UK listing already, so this will be the second location. We can address this after we get it live though, since we have a few details still to work out. Thanks.

dtothefp:
@kellyfinn logo works for me. As you mentioned looks kind of crappy with two of the same logos next to one another

screen shot 2015-04-16 at 11 11 30 am

PR #984, opened 2015-04-10
remove tr helper from pricing modals

dtothefp:
This PR addresses `tr` helper not removing backslash escape characters that is messing with translation strings and causing bad UX.

#### To Test

Go to /pricing and inspect 

- `id="pricing-plan-signup-thank-you-cont"`
- `id="downgrade-plan-cont"`

And make sure there are no remaining `\` characters.

image 2


stewsmith:
LGTM

PR #992, opened 2015-04-12
Add webpack

dtothefp:
This PR addresses adding webpack module loader into the marketing repo.  I know this implementation isn't a very "webpack" way, but I'm first trying to mimic our build behavior that was happening with the `concat` task to get everything working, and then go back and refactor. 

#### Issues:
- [x] Build does not use `grunt-webpack` but rather the webpack API within a `grunt.registerTask`. This was much more advantageous for usage of env configs appended on the grunt object that the webpack task depends upon.
- [x] Add inline sourcemap support. Must add the `website-guts` directory to the [chrome dev tools workspace](https://developer.chrome.com/devtools/docs/workspaces).
- [x] Dev and production config, in prod Uglify and Dedupe, in dev add the Hot Module Replacement plugin in preparation to convert to Webpack Dev Server for dev livereload.
- [x] Bundling vendor files (along with global JS) into one file and making their globals available outside of the bundle.  Currently, I'm accomplishing this through the `script-loader`, which seems to concat all files into a string and call `eval` on it.  Libraries are required directly from Bower in `global.js` and bundles with all other global files.
- [x] Mimicking `concat` behavior for the `namespacePages` target https://github.com/optimizely/marketing-website/blob/master/grunt/concat.js. i.e. separate bundles will be created for `page` and `layout` level JS and they will be output to `dist/assets/js` directory in the appropriate directory structure `layout/*.js` or `page/*.js`.  
- [x] Create banner and footer wrapper using `inject-filename-loader` which injects a `targetName` variable for JS error logging of bundles in try/catch blocks in staging/prod.
- [ ] Handlebars is being bundled more than once.  Once in global.js and additional times in page/layout js. Find a way to have handlebars only bundles once, or bundled globally and not included in the hbs-loader requires. Was recommended the [Common Chunks Plugin](https://github.com/webpack/docs/wiki/list-of-plugins#user-content-commonschunkplugin) ( not merge blocking )
- [x] Move JSHint into webpack. All JSHint is happening through webpack except for the linting the test and server side files. JSHint config for all tasks is still necessary as it is `require`d in the `webpack` grunt task.
- [x] Om (PPC) pages not using webpack, still using concat. This is because these must support very old browsers and must maintain their separate build environment.

### To Test

make sure to `rm -rf node_modules && npm cache clear && npm i`
`rm -rf bower_components && bower cache clean && bower i`
run `grunt server`
- watch the build, that webpack compiles correctly. It will be halted on any JSHint errors.
- inspect your console to make sure there are no JS errors
- inspect the dist directory and see that all of the JS files have an associate inline source map and that individual modules are wrapped in the closure transforming window and document to `w` and `d`, respectively.
- test a page like `/events` that uses a client side handlebars template. These are now being compiled with the webpack handlebars loader.
- in the `grunt/webpack/index.js` file find the `opts` object.  Append a `console.log(targetName)` to the end of the query string 
```js
injectFileNameParams: [
        'inject=' + encodeURIComponent('var targetName = __filename.replace("website-guts/assets/js/", "");\n\nconsole.log(targetName);\n\n'),
        '&banner=' + encodeURIComponent(banner),
        '&footer=' + encodeURIComponent(footer)
      ].join('')
```
run the `grunt server` task again.  You will see all individual JS module file names logged.  This is what is used in error reporting to GA and debugging in production.

- check [Staging](https://www.optimizelystaging.com/dfoxpowell/webpack-integration/pricing/)
- look in `website-guts/assets/js/global.js` and see how the `require` functions bootstrap the JS in a particular order and omits certain files using a RegEx with `require.context`.  This order is important for internal dependencies
- play around with some es6 if you want
- throw an error in the JS and check that it maps back to the appropriate source file through the source map http://cl.ly/27173s2T3p00
- run grunt build and inspect that inline source maps are not present and that individual modules are wrapped in try/catch blocks and that code has been uglified
- run the tests: 
```
grunt server
grunt ui-test //in a separate tab

grunt test //run this instead of the two commands above
```

### Enhancements
- we can now write es6
- source maps for local JS debugging
- JSHint inside webpack
- elimination of handlebars task - handlebars compiled on the fly in webpack loader. Also webpack loader handlebars has the same syntax as Assemble v6 for handlebars helpers which will be helpful for @4vanger `tr` helper.
- drastic reduction of code in concat, JSHint, and uglify task
- Handlebars is no longer required in the vendor dependencies
- future elimination of globals substituted by `require`s
- JS files are watched by webpack which applies a cache. First build is slowish, but subsequent live changes with server running are fast.
- probably lots of other stuff I'm not thinking of

### Additional
- removed `yepnope` from `bower.json`. We deprecated this a long time ago.  `moment` was also removed and is now being `require`d from NPM


website-guts/assets/js/components/utility-nav.js
```diff
@@ -68,7 +69,7 @@ window.optly.mrkt.showUtilityNav = function (acctData, expData) {
 
     $('body').addClass('signed-in').removeClass('signed-out');
 
-    $('#signed-in-utility').html( window.optly.mrkt.templates.loggedInUtilityNav(handlebarsData) );
+    $('#signed-in-utility').html( loggedInUtilityNav(handlebarsData) );
```
stewsmith: love it

stewsmith:
On /pricing in staging the "Get Started Now" button isn't popping up a modal.

dtothefp:
@stewsmith the poor behavior was happening on that dialogin staging even before webpack. Remember you and I were working with that last week. Could you try again and confirm. Also helpful for an issue like this could you report if there are errors in the console, if network requests were successful and if you tried other browsers?

grunt/webpack/webpack.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = require('./make-webpack-config')({
```
4vanger: What is the reason to have this file - it is not used anywhere?

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,104 @@
+var _ = require('lodash');
```
4vanger: Why you separate this into separate file? After this config is spread over make-webpack-config.js, webpack/index.js and grunt/_webpack.js
If we will have everything in one file - it will be much easy to support it.

grunt/webpack/index.js
```diff
@@ -0,0 +1,120 @@
...
+      entry: entryBundle,
+      env: env,
+      outputPath: options.dest,
+      publicPath: options.publicPath,
+      jshintConfig: _.merge({}, jshintEnvConfig, serverJshintConfig),
+      injectFileNameParams: [
+        'inject=' + encodeURIComponent('var targetName = __filename.replace("website-guts/assets/js/", "");\n\n'),
+        '&banner=' + encodeURIComponent(banner),
+        '&footer=' + encodeURIComponent(footer)
+      ].join('')
```
4vanger: Array.join doesn't give any advantages here - just string concatenation will work faster and will be more readable

grunt/webpack/index.js
```diff
@@ -0,0 +1,120 @@
...
+     *   'layouts/seo': './website-guts/assets/js/layouts/seo.js',
+     *   'pages/android': './website-guts/assets/js/pagess/android.js'
+     * };
+     *
+     * because bundle name (i.e. key) has a slash it gets dynamically written to that direcotry
+     * by output: {filename: '[name].js'}
+     * Wrapping value in an array [ ] is a hack recommended by @sokra so multiple bundles
+     * that may `require` from one another do not throw errors in the build.
+     */
+    var expandPath = function(basePath, dirs) {
```
4vanger: Need a comment what this function does

grunt/webpack/index.js
```diff
@@ -0,0 +1,120 @@
...
+    var config = grunt.config.get('_webpack');
+    var options = config.options;
+    var env = options.environment;
+    var banner = options.banner;
+    var footer = options.footer;
+
+    var pages = expandPath(config.pages.cwd, config.pages.src);
+
+    var globalBundle = {
+      bundle: ['./' + config.globalBundle.cwd + config.globalBundle.src]
```
4vanger: Why that's the only place where globalBundle is used - why not to store it as concatenated path? Same for "pages" above.
And in general - why need to calculate all of this on every run and not put "entryBundle" with all the data into config object initially.

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,104 @@
...
+  var jshintConfig = _.merge({}, {
+    emitErrors: true,
+    failOnHint: true,
+  }, opts.jshintConfig);
+
+  var preloaders = [
+    {
+      test: /\.js$/,
+      exclude: [
+        /node\_modules/,
```
4vanger: no need to escape underscores in regexps

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,104 @@
...
+        /bower\_components/,
+        /libraries/,
+        /form\-filler/,
```
4vanger: no need to escape minus as well

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,104 @@
...
+  ];
+
+  var loaders = [
+    { test: /\.hbs$/, loader: 'handlebars-loader' },
+    {test: /\.js?$/, exclude: ['bower_components', 'node_modules'], loader: 'babel-loader'}
+  ];
+
+  var plugins = [
+    new webpack.ResolverPlugin(new webpack.ResolverPlugin.DirectoryDescriptionFilePlugin('bower.json', ['main'])),
+    new webpack.NoErrorsPlugin()
```
4vanger: Please add https://github.com/optimizely/l10n-tr-plugin into list as well

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  ];
+
+  var loaders = [
+    { test: /\.hbs$/, loader: 'handlebars-loader' },
+    {test: /\.js?$/, exclude: ['bower_components', 'node_modules'], loader: 'babel-loader'}
+  ];
+
+  var plugins = [
+    new webpack.ResolverPlugin(new webpack.ResolverPlugin.DirectoryDescriptionFilePlugin('bower.json', ['main'])),
```
4vanger: Comment with explanation will be good here

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,104 @@
...
+    {
+      test: /\.js$/,
+      exclude: [
+        /node\_modules/,
+        /bower\_components/,
+        /libraries/,
+        /form\-filler/,
+        /global\.js/
+      ],
+      loader: 'inject-filename-loader?' + opts.injectFileNameParams
```
4vanger: Comment will be good to explain what happens here

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,104 @@
...
+
+  //TODO: Add webpack dev server for Hot Module Replacment
+  var devPlugins = [
+    new webpack.HotModuleReplacementPlugin()
+  ];
+
+  var prodPlugins = [
+    new webpack.optimize.UglifyJsPlugin({output: {comments: false}}),
+    new webpack.optimize.DedupePlugin(),
+    new webpack.DefinePlugin({
```
4vanger: Comment with explanation will be good here

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,104 @@
...
+
+  if(opts.env === 'dev') {
+    plugins.push.apply(plugins, devPlugins);
+  } else {
+    plugins.push.apply(plugins, prodPlugins);
+  }
+
+  var webpackConfig = {
+    node: {
+      __filename: true
```
4vanger: Comment with explanation will be good here

package.json
```diff
@@ -48,18 +50,26 @@
     "grunt-text-replace": "0.3.12",
     "grunt-userevvd": "0.1.4",
     "gulp-extname": "0.2.0",
-    "handlebars": "3.0.0",
+    "gulp-util": "^3.0.4",
+    "handlebars-loader": "1.0.0",
+    "imports-loader": "0.6.3",
+    "inject-filename-loader": "git+ssh://git@github.com:dtothefp/inject-filename-loader.git",
```
4vanger: 1) why this repo is not under optimizely account? Jenkins won't be able to access it.
2) need to use explicit commit version - same for l10n-tools below - so we can be sure that latest changes won't break current version

website-guts/assets/js/components/modal-error.js
```diff
@@ -1,4 +1,4 @@
-var tr = w.optly.tr;
+var tr = require('translation');
```
4vanger: we need to start using l10n-tr repo and require it

website-guts/assets/js/global.js
```diff
@@ -1,624 +1,22 @@
...
-
-  var text = '';
-  var possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
-
-  for(var i=0; i < 8; i++) {
-    text += possible.charAt(Math.floor(Math.random() * possible.length));
+require('script!history.js/scripts/bundled/html4+html5/jquery.history');
+require('script!oform/dist/oForm.min');
+
+var allGlobals = [
```
4vanger: Comment with explanations will be really useful here

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+      }
+
+    });
+
+  }
+
+};
+
+window.optly.mrkt.activeLinks.markActiveLinks();
```
4vanger: why are you calling this function twice? first time at line 47

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+    result[item[0]] = item[1];
+
+  }
+
+  return result;
+
+};
+
+window.optly.mrkt.inlineFormLabels = function(){
```
4vanger: Why do we have 2 copies of exactly same function?

website-guts/assets/js/globals/browser.js
```diff
@@ -0,0 +1,125 @@
+var BrowserDetect = {
```
4vanger: It will be good to add comment here.

website-guts/assets/js/globals/browser.js
```diff
@@ -0,0 +1,125 @@
...
+    {string: navigator.userAgent, subString: 'Firefox', identity: 'Firefox'},
+    {string: navigator.userAgent, subString: 'Safari', identity: 'Safari'},
+    {string: navigator.userAgent, subString: 'Opera', identity: 'Opera'}
+  ]
+
+};
+
+BrowserDetect.init();
+
+window.optly.mrkt.browser = BrowserDetect.browser;
```
4vanger: Since we have webpack now - we can get rid of this global variables and just require everything we need

website-guts/assets/js/globals/browser.js
```diff
@@ -0,0 +1,125 @@
...
+
+  } else {
+
+    return false;
+
+  }
+
+};
+
+window.optly.mrkt.mobileJS = function(){
```
4vanger: comments?

website-guts/assets/js/globals/browser.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  var querystring = window.optly.mrkt.utils.deparam(window.location.href);
+
+  if( (querystring.site_mode && querystring.site_mode === 'mobile') ) {
+    $.cookie('optimizelySiteMode', 'mobile', { path: '/' });
+  }
+
+  var mobileNavBound = false;
+  $(window).on('load resize', function() {
```
4vanger: comments?

website-guts/assets/js/globals/change_plan.js
```diff
@@ -0,0 +1,180 @@
...
+       args.load (function) = when the response on the http request is received, receives the XMLHttpRequestProgressEvent as an argument
+       args.abort (function) = when the request is aborted, receives the XMLHttpRequestProgressEvent as an argument
+       args.error (function) = when there is an error, receives the XMLHttpRequestProgressEvent as an argument
+       args.callback (function) = a callback for the load event, gets passed the load event
+
+*/
+
+    if(typeof args.plan === 'string' && args.plan){
+
+      var setPlan = new XMLHttpRequest();
```
4vanger: why we don't use $.ajax here?

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+    });
+
+  }
+
+};
+
+window.optly.mrkt.activeLinks.markActiveLinks();
+
+w.optly.mrkt.formDataStringToObject = function getJsonFromUrl(string) {
```
4vanger: duplicate declaration

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
+window.optly.mrkt.anim= window.optly.mrkt.anim || {};
+
+window.optly.mrkt.formDataStringToObject = function getJsonFromUrl(string) {
```
4vanger: comments?

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+  var data, result, i;
+
+  data = string.split('&');
+
+  result = {};
+
+  for(i=0; i<data.length; i++) {
```
4vanger: data.reduce will be more readable

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+    var item = data[i].split('=');
+
+    result[item[0]] = item[1];
```
4vanger: don't we need to decodeURIComponent here? item[1] will be escaped

website-guts/assets/js/globals/browser.js
```diff
@@ -0,0 +1,125 @@
...
+
+BrowserDetect.init();
+
+window.optly.mrkt.browser = BrowserDetect.browser;
+
+window.optly.mrkt.browserVersion = BrowserDetect.version;
+
+window.optly.mrkt.isMobile = function(){
+
+  if( /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent) ) {
```
4vanger: can be simplified to return /.../.test(...)

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+  }
+
+  return result;
+
+};
+
+//apply active class to active links
+w.optly.mrkt.activeLinks = {};
```
4vanger: this variable is not used anywhere except this file - no need to put it into global variable

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+w.optly.mrkt.activeLinks.currentPath = w.location.pathname;
```
4vanger: this value is used twice few lines below - just put it into local variable - no need to put it into global namespace

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+w.optly.mrkt.activeLinks.markActiveLinks = function(){
```
4vanger: This function is used immediately after in this file - no need to put it into function

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+  $('a').each(function(){
+
+    if(
+
+      $(this).attr('href') === w.optly.mrkt.activeLinks.currentPath ||
```
4vanger: var href = $(this).attr('href');
if(w.location.pathname in [href, href + '/'])

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+    }
+
+  });
+
+};
+
+w.optly.mrkt.activeLinks.markActiveLinks();
+
+w.optly.mrkt.inlineFormLabels = function(){
```
4vanger: comment explaining function is missing

4vanger: Also same function is defined 2 more time in sources

website-guts/assets/js/globals/general.js
4vanger: 3 times - one more in assets/js/om/global.js

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+  var possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+
+  for(var i=0; i < 8; i++) {
+    text += possible.charAt(Math.floor(Math.random() * possible.length));
+  }
+
+  return text;
+};
+
+w.optly_q.push([function(){
```
4vanger: what's happening here? comment explaining this will be really useful

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+    if(typeof w.optly_q.acctData === 'object'){
+
+w.analytics.ready(function(){
```
4vanger: broken indentation

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+  }, 1000);
+}
+
+if( $.cookie('amplitude_idoptimizely.com') ) {
+  w.optly.mrkt.deleteCookie('amplitude_idoptimizely.com', { path: '/', expires: -5, domain: '.optimizely.com'} );
+}
+
+w.optly.mrkt.setAttributeCookie = function(signInResponse){
+
+  var setCookieValue = function(args){
```
4vanger: too many newlines in this function - makes it hard to read

website-guts/assets/js/globals/general.js
```diff
@@ -0,0 +1,277 @@
...
+
+    var userAtrributeValues = [],
+    planMap = w.optly.planMap,
+    planCode,
+    plan;
+
+    if(typeof args === 'object'){
```
4vanger: it will be good to explain what happens here

website-guts/assets/js/globals/general.js
4vanger: Why we need to keep this function in closure? it doesn't use any data from closure

website-guts/assets/js/utils/form_helper_factory.js
```diff
@@ -1,4 +1,4 @@
-var tr = w.optly.tr;
+var tr = require('translation');
```
4vanger: use l10n-tr repo here

grunt/webpack/index.js
```diff
@@ -0,0 +1,120 @@
...
+     * Wrapping value in an array [ ] is a hack recommended by @sokra so multiple bundles
+     * that may `require` from one another do not throw errors in the build.
+     */
+    var expandPath = function(basePath, dirs) {
+      return grunt.file.expand({cwd: basePath}, dirs).reduce(function(map, filepath) {
+        map[filepath.replace('.js', '')] = ['./' + path.join(basePath, filepath)];
+        return map;
+      }, {});
+    };
+    var config = grunt.config.get('_webpack');
```
4vanger: I wasn't able to understand what will be in config after this - does this includes grunt/_webpack.js?

grunt/_webpack.js
```diff
@@ -0,0 +1,27 @@
...
+      colors: true,
+      modules: true,
+      reasons: true
+    },
+    dest: '<%= config.dist %>/assets/js/'
+  },
+  pages: {
+    src: ['pages/*.js', 'layouts/*.js'],
+    cwd: '<%= config.guts %>/assets/js/',
+    dest: '<%= config.dist %>/assets/js/'
```
4vanger: as far as I see -  "dest" is never used

grunt/_webpack.js
```diff
@@ -0,0 +1,27 @@
...
+  },
+  globalBundle: {
+    src: 'global.js',
+    cwd: '<%= config.guts %>/assets/js/',
+    dest: '<%= config.dist %>/assets/js/'
```
4vanger: as far as I see -  "dest" is never used

4vanger:
webpack config is spread across grunt/_webpack.js, grunt/webpack/index.js,  grunt/webpack/make-webpack-config.js and grunt/webpack/webpack.config.js
It takes lots of time to understand dependencies between this files. this must be simplified

grunt/webpack/index.js
```diff
@@ -0,0 +1,120 @@
...
+
+    var jshintEnvConfig = env === 'dev' ? jshintConfig.clientDev.options : jshintConfig.clientProd.options;
+
+    var opts = {
+      entry: entryBundle,
+      env: env,
+      outputPath: options.dest,
+      publicPath: options.publicPath,
+      jshintConfig: _.merge({}, jshintEnvConfig, serverJshintConfig),
+      injectFileNameParams: [
```
4vanger: string is created here but used in different file. In case you would like to use makeConfig somewhere else - you will need to do this preparation (obj -> query string) at that file as well. So it is much better to send plain object here and convert it before usage.

grunt/webpack/webpack.config.js
dtothefp: @4vanger I had it there because if we ever ran with the webpack CLI it would be useful. But that will probably never happen so I will delete it

grunt/webpack/make-webpack-config.js
dtothefp: I like to keep the creation of the webpack config object separate from the actual grunt task. This is more inline with webpack philosophy for config file. Also in this scenario the actual `grunt.registerTask` will not be altered, only the `_webpack` file which is similar to a grunt `src` `dest` `options` task and in the `make-webpack.config`

grunt/webpack/index.js
dtothefp: @4vanger I don't understand this comment. This is only run once in the grunt task and not again in the watch. For multiple entry points webpack expects and object

```
     var entryObj = {
        'layouts/seo': ['./website-guts/assets/js/layouts/seo.js'],
        'pages/android': ['./website-guts/assets/js/pagess/android.js'],
        'bundle': ['./website-guts/assets/js/global.js']
      };

```

package.json
dtothefp: @4vanger good call, I just added it to optimizely repo

website-guts/assets/js/globals/general.js
dtothefp: @4vanger everything is `om` is a completely separate build. It's JS is compiled using `concat` because this is our PPC page that must have it's own environment and support old browsers

website-guts/assets/js/globals/browser.js
dtothefp: @4vanger agreed but that will be in a separate PR

website-guts/assets/js/globals/change_plan.js
dtothefp: @4vanger for some reason kyle did this with raw JS. I agree I'd rather use $.ajax but again that will be a separate PR

website-guts/assets/js/globals/general.js
dtothefp: good catch

website-guts/assets/js/globals/general.js
dtothefp: @4vanger I'm not sure if this is a function Kyle is using for testing but I don't see it being used anywhere so I can't really say about the use case

website-guts/assets/js/globals/general.js
dtothefp: @4vanger again this is Kyle's code...I agree with the newline issue

grunt/webpack/index.js
dtothefp: @4vanger not sure what you mean? Query String for `inject-filename-loader` is dependent on templating done by grunt, so this config must be resolved inside the grunt task and not `make-webpack.config`

grunt/webpack/index.js
jordangarcia: This seems a bit unnecessary for the global bundle since you aren't expanding anything.  Given that this is following the convention of the other bundles to be in a config file it seems fine.

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,124 @@
...
+        /bower_components/,
+        /libraries/,
+        /form-filler/,
+        /global\.js/
+      ],
+      loader: 'inject-filename-loader?' + opts.injectFileNameParams
+    },
+    {
+      test: /\.js$/,
+      exclude: [
```
jordangarcia: This doesnt seem like the most maintainable solution for exlcusion from a loader.  Dont know if any of the bottom files live in a the same place thought.

grunt/webpack/make-webpack-config.js
```diff
@@ -0,0 +1,124 @@
...
+  ];
+
+  /**
+   * Plugins specific to production
+   * -- Uglifiy JS and strip comments
+   * -- Deduplicates equal or similar files http://webpack.github.io/docs/list-of-plugins.html#dedupeplugin
+   * -- Defines a global production variable that can be used in JS if necessary
+  */
+  var prodPlugins = [
+    new webpack.optimize.UglifyJsPlugin({output: {comments: false}}),
```
jordangarcia: putting a `sourceMaps: false` option in UglifyJs speeds it up a lot.

website-guts/assets/js/global.js
```diff
@@ -1,624 +1,36 @@
...
-w.optly.tr = function(str) {
-  // If w.optlyDict is present - use it for dictionary lookup.
-  // Need to have a global variable here because it must be declared *before* app load because app may need to localize
-  // message during initialize.
-  if(typeof w.optlyDict !== 'undefined' && w.optlyDict[str] !== null) {
-    str = w.optlyDict[str];
+var allGlobals = [
+  require.context('./utils', false, /^(?!\.\/form-filler\.js).*\.js$/),
+  require.context('./utils/form_helpers', false, /^(?!\.\/form_helper_factory\.js).*\.js$/),
+  require.context('./globals', false, /^(?!\.\/translation\.js).*\.js$/),
```
jordangarcia: what is this regex doing?  is it excluding certain files?

dtothefp: @jordangarcia yeh I put some additional comments in about that at @4vanger request

https://github.com/optimizely/marketing-website/blob/dfoxpowell/webpack-integration/website-guts/assets/js/global.js

Basically excludes certain files that are order dependent or unnecessary in the build

grunt/webpack/make-webpack-config.js
dtothefp: cool I copied your config from optimizely repo

```
    new webpack.optimize.UglifyJsPlugin({
      output: {
        comments: false
      },
      compress: {
        conditionals: false,
        warnings: false,
      },
      sourceMap: false
    }),
```

stewsmith:
LGTM

PR #1018, opened 2015-04-17
add default seo title suffix

dtothefp:
This PR addresses https://optimizely.atlassian.net/browse/MKT-1696 requesting pages with a `TR_seo_title` key to have ` - Optimizely` appended to the value.

#### To Test

Find a page like /partners/solutions and inspect that the title in the browser tab got appended with ` - Optimizely`.

Now check a page with no `TR_seo_title` key such as the homepage and inspect that the title in the tab is `'Optimizely: Make every experience count'`


stewsmith:
LGTM

PR #1023, opened 2015-04-21
Update global_feature_list.yml

teeyoutan:
modifying for android launch


teeyoutan:
@KarenLoughrey  FYI

KarenLoughrey:
@teeyoutan Translations are in progress, will let you know as soon as they're ready (should be about 48 hrs).

stewsmith:
@KarenLoughrey I've updated the footer copy per [the ticket](https://optimizely.atlassian.net/browse/PMM-177). This may also need translation.

KarenLoughrey:
@stewsmith @teeyoutan Translations are done. Having a couple of issues with the iOS --> Mobile (translations are showing up on staging, but not on live pages), but I think we can go ahead and merge and I'll look into that .

stewsmith:
@KarenLoughrey Okay thanks. Deploying now.

PR #1043, opened 2015-04-28
Move roundtable discussion

stewsmith:
The Roundtable discussion on the Option Agenda needs to be on Day 1 of Opticon


danoc:
:+1:

PR #1045, opened 2015-04-29
drop message variants

4vanger:


grunt/assemble/plugins/translation-utils/translate-page-data.js
```diff
@@ -22,25 +23,13 @@ module.exports = function(assemble){
...
-      var childTrans = objParser.translate(childLang || {}, childTranslations);
-
-      _.merge(pageDataClone[locale][fp], parentTrans, childTrans);
+      var parentLang = (lang[websiteRoot] && lang[websiteRoot][parentPath]) || {};
+      var childLang = (lang[locale] && lang[locale][fp]) || {};
+      //order is important because child keys must overwrite parent
+      var combinedYml = _.merge({}, parentLang, childLang);
+      var trans = objParser.translate(combinedYml, translations);
+      fixHTMLEscape(trans);
+      _.merge(pageDataClone[locale][fp], trans);
```
dtothefp: @4vanger adjusted this function to account for external YML in locale subfolder that has no template. In this case the YML keys should overwrite the parent

screen shot 2015-04-29 at 12 08 31 pm

PR #1046, opened 2015-04-29
Opticon Speakers Page

stewsmith:


website-guts/assets/js/components/modals.js
```diff
@@ -147,9 +148,15 @@ window.optly.mrkt.modal.open = function(modalArgs) {
...
 
-  $('html, body').addClass('modal-open');
-
-  window.scrollTo(0,0);
+  var isMobile = w.optly.mrkt.isMobile();
+  if (modalPosition) {
+    if (!isMobile) {
+      $elm.find('.dialog').css('top', modalPosition);
+    }
+  } else {
```
dtothefp: @stewsmith this position can actually be done better in css 

```css
.modal-dialog-container .dialog  {
position: fixed;
top: 50%;
left: 50%;
transform: translate(-50%, -50%);
margin-top: 0;
}
```

Obviously you would put this in a media query so it doesn't affect mobile. 

also the modal will now not scroll off the page. I think you should change the prop from `modalArgs.position` to something like `modalArgs.static` or whatever and then do 

```js
if(!static) {
$('html, body').addClass('modal-open');
 window.scrollTo(0,0);
}

```

stewsmith: Oh yeah that is such a better solution.

dtothefp:
@stewsmith LGTM

PR #1067, opened 2015-05-06
Consolidate colors and brand color variables

julesforrest:
@stewsmith Color consolidation of marketing site


dtothefp:
@julesforrest look above at the `hound` CSS linter. Can you fix these issues?

julesforrest:
@dtothefp Stew told me to ignore some of the comments that refer to older files that followed a different pattern (I was just swapping color values so all the spacing issues are based on the existing tabs and spacing), but I can clean this up when I get into the office

stewsmith:
@dtothefp @julesforrest Most of the comments are indentation which we should fix all at one time. There's another comment about color literals, which we can ignore because it's in a linear gradient.

@julesforrest I would just add a space after the commas.

dtothefp:
@julesforrest @stewsmith there should be an easy way in the text editor to switch from tabs to spaces. In vim you could just highlight and hit `=`

stewsmith:
LGTM :package:

PR #1088, opened 2015-05-11
Add descriptions for blue acorn and decibel

stewsmith:
This adds descriptions on the modals for blue acorn and decibel


stewsmith:
LGTM to myself

PR #1094, opened 2015-05-12
update webpack config for Handlebars includePaths, Common Chunks and JSHint

dtothefp:
- [x] add include paths for client side hbs helpers
- [x] get CommonsChunksPlugin working for bundles that contain common libs such as HBS otherwise it is bundled twice
- [x] fix jshint in webpack

## To Test
https://www.optimizelystaging.com/dfoxpowell/upgrade-webpack-config/
- checkout the `{{debug thing_to_log}}` client side helper. Put it in any client side hbs template, navigate to the page, and check the console. The helpers for client side live in `website-guts/client-helpers/`
- CommonsChunksPlugin creates a bundle with common code shared between entry bundles. In this case the global bundle and some page level bundles have client side `hbs` templates. Therefore in order to avoid bundling handlebars twice on a page it needs to be chunked out. This will be a good solution moving forward for things like `Vue` as well, and they can be added in the `_webpack` config as `commonVendorScripts`. Navigate to the home page and login. Check that the logged in utility nav works. Go to the jobs page and make sure that client side template works. Inspect the `bundle.js` and `jobs.js` and see that `Handlebars` is not bundled in. Check the `commons.js` and see that handlebars is bundled
- For jshint in dev first compile all JS. Then add something bad like `asdfasdf` to a file and save. Check that jshint catched the error. Check that things like `debugger` and `console.log` don't throw an error. Now add a `debugger` and `console.log` in some file and run `grunt config:production webpack` and ensure that this WILL throw an error.


stewsmith:
Followed your testing script, everything works.

screen shot 2015-05-21 at 6 11 38 pm
Are these warnings expected? If so, PR gets LGTM.

dtothefp:
@stewsmith those warnings are expected. Something to do with chunking the Handlebars library.

PR #1096, opened 2015-05-12
add locale tracking data to oform middleware

dtothefp:
@arthuracs @4vanger 
Adds temporary tracking of language key for `/account/create` based upon `ccTLD`.

- only works for requests made to `/account/create` (ie no `/account/free_trial_create`. @arthuracs let me know if this should be added as well but it would mean updating another endpoint.

## To Test
- go to any page that uses `/account/create` i.e. homepage, mobile page
- preserve the log in chrome network and inspect the request payload sent
- should see a `locale` key
- default is `us_EN`, unfortunately we cannot test this in staging for `ccTLD` unless you have some good ideas of how.

staging url: https://www.optimizelystaging.com/dfoxpowell/language-selection-cctld/

screen shot 2015-05-13 at 10 07 47 am

### DIFF
--- a/website-guts/assets/js/utils/no-global/get-language-key.js
+++ b/website-guts/assets/js/utils/no-global/get-language-key.js
@@ -1,0 +1,26 @@
+/**
+ * Utility function to append locale specific data to oForm payload object sent to /account/create endpoint
+ *
+ * @param {String} `data` string of form data encoded by oForm
+ * @return {String} encoded locale string to be appended to the oForm `data` string in the oForm middleware
+ */
+
+module.exports = function(data) {
+  var langKey;
+  var tld = window.location.hostname.split('.').pop();
+
+  if(window.optly.l10n && window.optly.l10n.locales) {
+    for(var key in window.optly.l10n.locales) {
+      if(key === tld) {
+        langKey = window.optly.l10n.locales[key];
+        break;
+      }
+    }
+
+    if(!langKey) {
+      langKey = 'us_EN';
+    }
+  }
+
+  return langKey ? window.encodeURI('&locale=' + langKey) : '';
+};
--- a/website-guts/assets/js/utils/no-global/get-language-key.js
+++ b/website-guts/assets/js/utils/no-global/get-language-key.js
@@ -6,8 +6,9 @@
  */
 
 module.exports = function() {
-  var langKey;
+  var DEFAULT_KEY = 'en_US';
   var tld = window.location.hostname.split('.').pop();
+  var langKey;
 
   if(window.optly.l10n && window.optly.l10n.locales) {
     for(var key in window.optly.l10n.locales) {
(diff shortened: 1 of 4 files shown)


dtothefp:
retest this please

website-guts/assets/js/utils/no-global/get-language-key.js
```diff
@@ -0,0 +1,26 @@
...
+  if(window.optly.l10n && window.optly.l10n.locales) {
+    for(var key in window.optly.l10n.locales) {
+      if(key === tld) {
+        langKey = window.optly.l10n.locales[key];
+        break;
+      }
+    }
+
+    if(!langKey) {
+      langKey = 'us_EN';
```
4vanger: it is "en_US"! Extremely important to change - otherwise request will fail

website-guts/assets/js/utils/no-global/get-language-key.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+  }
+
+  return langKey ? window.encodeURI('&locale=' + langKey) : '';
```
4vanger: we always have a langKey - no need to do this check

website-guts/assets/js/utils/no-global/get-language-key.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Utility function to append locale specific data to oForm payload object sent to /account/create endpoint
+ *
+ * @param {String} `data` string of form data encoded by oForm
+ * @return {String} encoded locale string to be appended to the oForm `data` string in the oForm middleware
+ */
+
+module.exports = function() {
+  var langKey;
```
4vanger: As for me we can do it more simple since we will be removing this soon but it is up to you to decide:
var langKey = {
  'www.optimizely.de': 'de_DE',
  'www.optimizely.de': 'de_DE',
  'www.optimizely.fr': 'fr_FR',
  'www.optimizely.jp': 'ja_JP,
  'www.optimizely.es': 'es_ES'
}[window.location.host]

website-guts/assets/js/utils/no-global/get-language-key.js
```diff
@@ -0,0 +1,26 @@
...
+
+  if(window.optly.l10n && window.optly.l10n.locales) {
+    for(var key in window.optly.l10n.locales) {
+      if(key === tld) {
+        langKey = window.optly.l10n.locales[key];
+        break;
+      }
+    }
+
+    if(!langKey) {
```
4vanger: need to do this check outside if() statement to always set default locale

4vanger:
Please see my comments - at least you MUST fix default locale value - current one will fail every request

website-guts/assets/js/utils/no-global/get-language-key.js
dtothefp: @4vanger I updated this, see above screenshot

website-guts/assets/js/utils/no-global/get-language-key.js
dtothefp: @4vanger moved check outside of `if` block

dtothefp:
@4vanger changed default lang key to `en_US` and moved check for `langKey` outside of the `if` block

dtothefp:
retest this please

stewsmith:
LGTM

PR #1113, opened 2015-05-19
Add robots.txt to marketing website

arthuracs:
This is for https://optimizely.atlassian.net/browse/MKT-1712

@stewsmith can you review? I tested this locally and it works at http://localhost:9000/dist/robots.txt but not sure if it's the correct place to add this.

cc: @jrietveld @ihsekat


dtothefp:
@arthuracs I think this shouldn't go in the assets directory but should just be in the root of `website` and we can copy it from there

stewsmith:
@arthuracs I agree with @dtothefp

arthuracs:
@dtothefp @stewsmith ok, no problem. Like this?

ihsekat:
LGTM, what do you think @stewsmith?

stewsmith:
LGTM

PR #1117, opened 2015-05-20
Fix padding on opticon titles

stewsmith:
The padding on the opticon titles is a bit off. This PR fixes that.


danoc:
Sweet. Go, Hound! LGTM.

PR #1118, opened 2015-05-20
Implement SSO on marketing website

danoc:
@stewsmith – This isn't good to merge yet, but can you take a look at what I have so far?

Still needs more design work.

**How to test:** On staging, click on "Log in", click on "Use single-sign on," then enter an email address. It should send an HTTP request and you should receive a response back saying that either there is no account with that email address or SSO is not enabled for that account.


danoc:
This still needs to handle form validation and AJAX responses, but @stewsmith, can you take a first pass?

website-guts/templates/components/modals/signin_modal.hbs
```diff
@@ -21,6 +21,14 @@ api_domain: true
                   <span class="password-related error-hide error-right">{{tr 'Password is required'}}</span>
                   <input type="password" name="password" required>
               </label>
+
+              <!--
```
stewsmith: What is this comment?

website-guts/assets/js/components/modals.js
```diff
@@ -116,6 +116,7 @@ function storeModalState(modalType, modalOpen) {
 }
 
 window.optly.mrkt.modal.open = function(modalArgs) {
+
```
stewsmith: why whitespace?

stewsmith:
LGTM

PR #1120, opened 2015-05-20
MKT-1945 -- Acceptable Use Policy Page

stewsmith:
https://optimizely.atlassian.net/browse/MKT-1945

Legal needs some terms under the url `/acceptable-use-policy`


website/acceptable-use-policy/index.hbs
```diff
@@ -0,0 +1,14 @@
...
+TR_visible_title: Acceptable Use Policy
+---
+<div class="row">
+  <div class="columns ten offset-one">
+    {{#with page_data}}
+    {{#each terms}}
+    <p>{{paragraph}}</p>
+    {{/each}}
+    {{/with}}
+  </div>
```
dtothefp: @stewsmith you can get rid of the `TR_paragraph` key an just have strings in the YML array. Then in the each loop just do `<p>{{this}}</p>`

dtothefp:
@stewsmith LGTM

PR #1129, opened 2015-05-26
CD-866 -- Add location filter to jobs page

stewsmith:
This PR adds a dropdown to the jobs page to filter jobs based on location.
https://optimizely.atlassian.net/browse/CD-866

@jules is handling the intro section and other styling.


website-guts/assets/js/pages/jobs.js
```diff
@@ -23,25 +23,93 @@ window.optly.mrkt.jobsPage.testimonials = function() {
 };
 
 $('#view-all-jobs').click(function() {
-    $('html, body').animate({scrollTop: $('#jobs-list').offset().top}, 700);
-    return false;
+  $('html, body').animate({scrollTop: $('#jobs-list').offset().top}, 700);
```
chestone: Nit: Space the object arg over a few lines. This reads funny.

website-guts/assets/js/pages/jobs.js
```diff
@@ -23,25 +23,93 @@ window.optly.mrkt.jobsPage.testimonials = function() {
...
 });
 
 window.optly.mrkt.jobsPage.testimonials();
 
+/*
+* Jobs and Department titles are hidden or
+* shown based on the filter criteria.
+*/
+function showOrHideJobs(filterText, $jobLocation, $closestDepartmentTitle, $departmentTitles) {
+  if (!filterText.includes($jobLocation.text())) {
```
chestone: String.includes has really limited support are we shimming this?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

Could rewrite this as a regex match `filterText.matches(new Regexp($jobLocation.text(), 'g')).length > 0` something like that.

website-guts/assets/js/pages/jobs.js
```diff
@@ -23,25 +23,93 @@ window.optly.mrkt.jobsPage.testimonials = function() {
...
+    if (filterText === 'All Locations') {
+      $closestDepartmentTitle.show();
+      $departmentTitles.show();
+    } else {
+      $closestDepartmentTitle.hide();
+      $departmentTitles.each(function() {
```
chestone: Using `this` to refer to the current iterable is a little confusing here since you are inside both an .each and then a .filter, you could name the parameter to avoid this and make it a bit more clear.

dtothefp:
@stewsmith @chestone I know I'm not a part of this anymore but I mentioned to Stew previously that this would be a nice use case for a first `Vue` component now that webpack is in place. The last thing the site needs is more jQuery spaghetti

stewsmith:
@dtothefp TBH I forgot about that discussion for this page. My fault. I don't want to rewrite this code with `Vue` now as the functionality is already done.

website-guts/assets/js/pages/jobs.js
stewsmith: Good catch on `includes`. I thought it was being shimmed but it wasn't. I think `indexOf` would be better in this case.

website-guts/assets/js/pages/jobs.js
```diff
@@ -23,25 +23,100 @@ window.optly.mrkt.jobsPage.testimonials = function() {
...
+    $('.job-location').each(function() {
+      var $jobLocation = $(this);
+      var $closestDepartmentTitle = $jobLocation.closest('li');
+
+      showOrHideJobs(filterText, $jobLocation, $closestDepartmentTitle, $departmentTitles);
+    });
+  });
+}
+
+function onlyUnique(value, index, self) {
```
chestone: Probably throw a comment on this as out of context I don't know what it's purpose is.

stewsmith: Good call.

chestone:
LGTM! Wish we could cleanup a bit of the jquery craziness, but not worth refactoring now. Also it got woof'd so I'm good :+1:

PR #1136, opened 2015-05-29
pull forked packages from optimizely GH

dtothefp:
This PR addresses pulling forked and custom packages from the Optimizely GitHub rather than individual GH users profile.  It also moves oform from bower to package.json pulling from Optimizely GH.

## To Test
`rmall` or
`nvm use 0.10 && rm -rf node_modules && npm cache clear && npm i`
`rm -rf bower_components && bower cache clean && bower i`
`grunt build`
`grunt connect:keepalive`

and check in staging

https://www.optimizelystaging.com/dfoxpowell/update-forked-dependencies

@stewsmith please review, @eric-optimizely this addresses what was discussed yesterday

@stewsmith updated this to pull globals with the `expose-loader` rather than the `script-loader` (this stringifies modules which is not nice because have to include minified version even in dev). `expose-loader` exposes a global to the window for non-commonjs module, but will allow us to include un-minified libraries in dev.

Check out shimming here:
http://webpack.github.io/docs/shimming-modules.html


website-guts/assets/js/global.js
```diff
@@ -1,5 +1,5 @@
 require('script!history.js/scripts/bundled/html4+html5/jquery.history');
-require('script!oform/dist/oForm.min');
+require('script!oForm/dist/oForm.min');
```
stewsmith: why is this camelCase now?

stewsmith:
@dtothefp thanks for moving these.

dtothefp:
@stewsmith `oForm` is camelcased in the script loader because that is how Kyle named it in his `package.json` so that is what Webpack will look for when it is `require`d

stewsmith:
LGTM

yesplease:
LGTM too

PR #1143, opened 2015-06-01
Speakers page and agenda updates

yesplease:
Removed Matty from agenda
Removed Deepa from agenda and from speakers page


stewsmith:
LGTM

PR #1148, opened 2015-06-02
Add sold out svg over partnership and enablement session info

yesplease:


stewsmith:
Code LGTM. Could you verify staging with the owners?

PR #1158, opened 2015-06-04
MKT-1946 -- Mobile redirect

stewsmith:
Redirect back to the mobile editor when referred from there.

https://optimizely.atlassian.net/browse/MKT-1946


mikeng13:
Ship it!

PR #1161, opened 2015-06-04
Disable Get Started Now button after user clicks once

yesplease:
Console log left in for testing on staging, remove before pushing live


stewsmith:
LGTM

PR #1173, opened 2015-06-09
Update Goldbach Interactive region

kellyfinn:


kellyfinn:
@stewsmith Can you take a look? This Solutions Partner isn't currently appearing when user sorts for region "EMEA - Germany". I tried changing to just "Germany" here but it still isn't appearing in the filter  so I changed it back to "EMEA - Germany". Thanks for the help!

stewsmith:
@kellyfinn I can reproduce the issue but I don't immediately see the cause. Can you make a ticket for this?

PR #1183, opened 2015-06-11
Rename /jobs, /pricing and /customers/customer-stories

stewsmith:
These pages will now live on /careers, /plans, and /case-studies (respectively)

@KarenLoughrey This is now ready for translation.


KarenLoughrey:
@stewsmith We had all these terms in the translation memory...so, translations are ready!

PR #1191, opened 2015-06-13
Bheinz/enable solutions forms - testing

yesplease:


stewsmith:
Merging to test leads going through on prod.

PR #1204, opened 2015-06-17
Remove homepage opticon banner html and css

stewsmith:


yesplease:
LGTM, staging looks good too

PR #1211, opened 2015-06-22
do not use app-wide locales property to find locale

4vanger:
INTL-376 Fix create account flow so that new user account language preference matches the Marketing Site they were using

### DIFF
--- a/website-guts/assets/js/utils/no-global/get-language-key.js
+++ b/website-guts/assets/js/utils/no-global/get-language-key.js
@@ -4,12 +4,6 @@
  * @param {String} `data` string of form data encoded by oForm
  * @return {String} encoded locale string to be appended to the oForm `data` string in the oForm middleware
  */
-module.exports = function() {
-  // extract TLD from current URL
-  var tld = window.location.hostname.split('.').pop();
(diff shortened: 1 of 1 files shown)


stewsmith:
@4vanger Is there a way I can test this?

4vanger:
I don't think so :( at least check that it returns en_US in staging env

stewsmith:
@4vanger How would I do that?

4vanger:
deploy app on staging (don't really know how to do that) and check that during registration it sends locale=en_US

arthuracs:
@stewsmith I tried going to www.optimizelystaging.com/andrew/locale-mapping but couldn't get the signup flow to work. Do you know if that should be possible on staging? Otherwise, can you verify this branch locally? It's a really bad customer experience right now.

stewsmith:
@4vanger @arthuracs I pulled this down and got a js hint error. I believe this was the reason the JS was not building correctly. I've moved the function declaration around to solve the error, but you can change it as you see fit.

Below is a screenshot of the form data going through. If this looks good to you then it LGTM and you can merge.

screen shot 2015-06-24 at 5 28 34 pm

PR #1214, opened 2015-06-22
Add back partners solutions js

stewsmith:
The js for partners/solutions was deleted so the dropdown isn't working anymore.

This PR adds the js back with a new file name.


yesplease:
Looks
Good
To
Me
:100:

PR #1215, opened 2015-06-22
TR_ on partners fix

stewsmith:
Right now some of the partners are not sorting correctly. This PR fixes that by prefixing `locations`, `region`, `state`, `city`, and `country` with`TR_`

To test, go to production and sort by "EMEA - Germany". Notice goldbach-interactive is not there. Try the same thing on staging and notice goldbach-interactive is there.


yesplease:
LGTM

PR #1218, opened 2015-06-23
Remove hardcoded references to optimizely.com

arthuracs:
to make sure links correctly resolve on international domains.

@stewsmith can you review? The references to www.optimizely.com jump visitors on international domains back to the English site. If we just make it a relative link, they'll stay on the international site. I fixed it in the places I could find, but please make sure to use relative links for any internal links.

@KarenLoughrey this should address the link issues you were seeing on https://www.optimizely.de/solutions/ and other international domains.

### DIFF
--- a/website/features-and-plans/global_features.yml
+++ b/website/features-and-plans/global_features.yml
@@ -116,9 +116,9 @@ enterprise_section:
   title: "Enterprise"
   subtitle: "Customized plans for your business."
   cta: "Talk to Us"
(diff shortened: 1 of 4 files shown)


arthuracs:
https://optimizely.atlassian.net/browse/INTL-378

stewsmith:
@arthuracs Ideally we would be able to use `linkPath` in `.yml` files but as of now, we can only use it in `.hbs` files. `linkPath` allows you to stay locally, on staging, or on your current domain.

That being said, using the relative path here is a better solution than what is currently live. LGTM.

PR #1219, opened 2015-06-23
Cdee/update grunt aws plugin

Chris-Dee:
Update since last attempt:

-- Greatly increased speed of upload, decreasing downtime (uploadConcurrency option)
-- Added fields that defaulted in grunt_aws but not in aws_s3 ( root dest)
-- Changed delete to be differential (Only deletes files not in cwd, so we don't have an empty bucket)

To test: Because the behavior of staging was different than prod in the last merge attempt, I would initially merge with the debug:true option on the production delete, so we could make sure that the upload won't fail before we actually delete the bucket (with the differential delete this shouldn't be an issue though, as the full bucket won't be deleted).


brianlucas:
This seems to do what we want.  Any concerns with merging it?  Looks good to me.

Chris-Dee:
Not really, just a little cautious since my last merge on this issue brought down the homepage. Now that all the options are the same for prod and staging (and it works on staging), there's no reason the upload should fail.

PR #1224, opened 2015-06-24
Hit greenhouse api for office data

stewsmith:
This PR fixes the "Amsterdam, Paris" in the jobs location filter. It pulls from a separate offices greenhouse endpoint.


yesplease:
LGTM

PR #1234, opened 2015-06-26
MKT-2075 -- Add swap tag to tech partners signup page

stewsmith:
Add a swap tag and id to /partners/technology/join so a section doesn't show up in japan.

https://optimizely.atlassian.net/browse/MKT-2075

To test: go to /partners/technology/join and inspect the last benefit. It should have a class of `SL_swap` and id of `JP_tech_join`

screen shot 2015-06-26 at 1 52 17 pm


yesplease:
LGTM

PR #1240, opened 2015-06-30
Updated sitemap.xml to have changefreq and priority.

bhamodi:
@stewsmith - This PR updates the sitemap to have `changefreq` and `priority` attributes. This helps SEO as the pages will get indexed more frequently.


stewsmith:
@ihsekat Can you review?

ihsekat:
Hey @bhamodi,

Thanks for taking a look at our sitemaps, but I'm not sure the new markup will be very helpful for SEO.

`changefreq` and `priority` are used to help Google determine crawl order for pages on our site. This can be helpful for larger sites that aren't being crawled and indexed properly. However, since our site is relatively small (~500 pages) Google is already crawling every one of our pages on a daily basis, so additional markup in the sitemap is superfluous.

Also, setting the same `changefreq` and `priority` to identical values for every page in the sitemap is not helpful, since that essentially gives every site on our site the same priority, which isn't the case. The marketing website also isn't updated on daily basis, so setting the `changefreq` to "daily" is inaccurate and could confuse Googlebot.

TLDR: We don't need to add additional markup to the sitemap since there are no crawling issues, especially if you're just giving every page on the site the same priority.

bhamodi:
Hey @ihsekat,

I'd recommend you check out this article, talking about how both those optional attributes are indeed useful for SEO purposes: http://www.v9seo.com/blog/2011/12/27/sitemap-xml-why-changefreq-priority-are-important/

Now for the specifics; the reason for having `changefreq` set to daily is because we have multiple daily deploys, and this I find to be an accurate representation.

As for `priority`, the default priority is actually set to `0.50` if not provided. I assigned all root domains (optimizely.com, optimizely.fr, optimizely.es, etc.) a priority of `1.00`. And all subdomains received a priority of `0.80`. This is a common technique. Google does it too except they use `0.50` for all their subdomains. (http://www.google.com/sitemap.xml) We can play with this number.

We don't necessarily have to have both, but in my opinion, it seems like having both would be useful. Thoughts?

ihsekat:
Hey @bhamodi, what is the benefit you expect to see from this change? Our entire site is already crawled on a daily basis, so explicitly changing `changefreq` to daily isn't going to change anything. 

Likewise explicitly setting crawl priority is not really necessary as Google is going to place more priority on the root domains anyway, since they are the pages that have the most backlinks.

Unless we are experiencing crawling or indexation issues, it's best to let Google figure these things out automatically.

bhamodi:
@ihsekat - I'm fine with closing this PR if it's true that we are get indexed everyday by at least the top 5 search engines in the market. Out of curiosity, how are we tracking who is indexing us and when we are being indexed? I'd be interested in finding out how we know that we're being indexed daily already. As for priority, I believe you are correct that the search engines should be able to determine appropriate priority themselves.

ihsekat:
@bhamodi - Google Search Console shows that Google is crawling all of our pages being crawled and that the pages in our sitemap are being indexed. If there is a change that we want indexed immediately, we can also submit those pages through Google Search Console.

bhamodi:
Cool, thanks @ihsekat!

PR #1242, opened 2015-06-30
fix free trial page logos

stewsmith:
On https://www.optimizely.de/free-trial/ The page looks like this:

screen shot 2015-06-30 at 4 47 24 pm

This was not visible on .com because there is an experiment running there.

After this change the page looks like this:

screen shot 2015-06-30 at 4 49 04 pm

I've also deleted the `modal-becom-partner-technology.js` file


yesplease:
LGTM

PR #1244, opened 2015-07-01
Add reporting to tab click on solutions

stewsmith:
On /solutions when the user clicks on one of these:
screen shot 2015-07-01 at 10 27 35 am
we need to track it.

I've verified this is going through to segment:
screen shot 2015-07-01 at 10 45 30 am


yesplease:
LGTM

PR #1255, opened 2015-07-07
Disable multiple submissions on form

yesplease:


yesplease:
@elliotykim could you please review?

elliotykim:
Hey @yesplease 

oforms can be submitted multiple times? Then the ideal solution would be to patch oform since this bug will affect every form using that plugin. Of course, it will require additional testing to make sure we don't break anything else.

Another solution is to use oform's "before" hook: https://github.com/optimizely/oform#before

You can move your logic here:
https://github.com/optimizely/marketing-website/blob/master/website-guts/assets/js/layouts/trial-form.js#L91

This is preferable to the current proposal b/c all of the form logic will be close together in the code. We like that: for maintainability and debugging purposes. Otherwise, we get sprinklings of logic here and there that can be time consuming to track down in the future.

yesplease:
Thanks @elliotykim! Will try moving logic into 'before' on oform and test it out. I'll do some extensive testing and resubmit for review.

yesplease:
@elliotykim 
Thanks for the feedback, Elliot. One of our projects for Q3 is to replace oForm so I opted to fix just this form.

You'll see I have moved the logic and added comments.
-Using oForm's 'before' method, updated code disables the submit functionality of the form
-If there is a validation error on the form, the form is re-enabled

Thanks for your review

elliotykim:
@yesplease 

That should do the trick. LGTM

PR #1264, opened 2015-07-13
Update Japanese URLs to .co.jp

ihsekat:
Updated all optimizely.jp URLs to optimizely.co.jp


ihsekat:
@yesplease @stewsmith Could you review? Thanks!

PR #1275, opened 2015-07-17
Jessie wants the social proof to be removed

jrietveld:
The number of experiments run is very far off reality plus we're not communicating this anywhere else on the site. Jessie asked for it to be removed. Currently we don't see this on the US website due to an experiment running on that page - but the intl sites are showing the original with this data.


yesplease:
Looks good to me but I will defer to @stewsmith to double-check since I think he built this page

stewsmith:
LGTM.

PR #1285, opened 2015-07-22
Remove trailing dash in yaml to fix table and make it display

yesplease:
A bug was introduced to this table in March. This PR fixes formatting on the YAML so handlebars can render that table again


stewsmith:
lgtm

PR #1289, opened 2015-07-24
Wrap slides URLs in quotes for consistency

yesplease:
Stew, as discussed, this is making our yaml more consistent by wrapping slides_url in quotes for each video


stewsmith:
lgtm

PR #1290, opened 2015-07-27
Update social share text on careers page

yesplease:
currently the default text for a careers page is generic. this is a great place for us to have a nicer message: we're reaching out to potential employees after all.


stewsmith:
lgtm

PR #1291, opened 2015-07-27
add event_ops to reporting object

stewsmith:


bhamodi:
I think this LGTM @stewsmith. But I'm pretty unfamiliar with this part of the codebase.

PR #1300, opened 2015-08-02
Fix local environment by upgrading to node v0.10.40

bhamodi:
@Chris-Dee and I have spent numerous hours investigating local environment issues and they all seemed to be resolved with this node version change.

`grunt server` now works again locally. The next step for us is to update the staging and production servers to have `v0.10.40` of Node installed.

Stay tuned @stewsmith and @yesplease.


bhamodi:
Closing this PR as we managed to get it working with the older version of node.

PR #1308, opened 2015-08-04
add apple-touch-icon.png to imgs and img copy task

stewsmith:
Adds `apple-touch-icon.png` to reduce 404s


bhamodi:
Thanks for doing this @stewsmith. LGTM.
One thing to think about is if we need to have multiple sizes for all the iOS versions? See this stackoverflow answer here: http://stackoverflow.com/a/21144916/3474514

I'm okay with shipping this first and observing logging to see if there are devices requesting certain sizes.

PR #1309, opened 2015-08-05
Bheinz/add opti glossary

yesplease:
Building out this ticket:
https://optimizely.atlassian.net/browse/MKT-2175

@elliotykim would you please review at your convenience? I think this page should work regardless of language (except for Japanese at this point). Gracias

@KarenLoughrey There may be some changes to the larger structure and script to this new Optimization Glossary page, but the human-readable content (all of which is contained in glossary.yml) should be finalized for the most part. There are a few potential tweaks, such as "AB" becoming "A/B" (pending feedback from @srusonis). Should we wait to start translating, or get started now and make the tweaks as needed? Thanks!

### DIFF
--- a/website-guts/assets/css/pages/glossary.scss
+++ b/website-guts/assets/css/pages/glossary.scss
@@ -1,0 +1,17 @@
+body.glossary {
+
+} 
+
+@media only screen and (min-width: $tablet-min-width) {
+  body.glossary {
+    .glossary-container {
+      max-width: 800px; 
+      margin: auto;
+    }
+    .term-container {
+      padding-bottom: 16px;
+    }
+
+
+  }
+}
--- a/website-guts/assets/css/pages/glossary.scss
+++ b/website-guts/assets/css/pages/glossary.scss
@@ -1,5 +1,11 @@
 body.glossary {
-
+  h1 {
+    color: $brand-blue;
+  }
+  .toc-index {
+    color: $dark-gray-text;
+    font-size: 22px;
+  }
 } 
 
 @media only screen and (min-width: $tablet-min-width) {
@@ -11,7 +17,5 @@ body.glossary {
     .term-container {
       padding-bottom: 16px;
     }
-
-
   }
 }
--- a/website-guts/assets/css/styles.scss
+++ b/website-guts/assets/css/styles.scss
@@ -33,6 +33,7 @@
 @import 'pages/events';
 @import 'pages/faq';
 @import 'pages/free-trial';
+@import 'pages/glossary';
 @import 'pages/getting-started';
 @import 'pages/homepage';
 @import 'pages/integrations-solutions';
--- a/website-guts/assets/js/pages/glossary.js
+++ b/website-guts/assets/js/pages/glossary.js
@@ -1,0 +1,76 @@
+$(function(){
+  //first gather all terms that are on page
+  var termsElements = $('.term-container');
+  
+  //build object with term and summary:
+  //key is string, value is the html <p> definition of term
+  var terms = {};
+  for (var i = 0; i < termsElements.length; i++){
+    var term = $.text(termsElements[i].children[0]);
+    var definition = termsElements[i].children[1];
+    terms[term] = definition;
+  }
+
+  //now we need to sort terms. objects are by definition unsorted, so let's grab all terms in an array
+  //and sort them in a case-insensitive way
+  var sortedTermsList = Object.keys(terms).sort(function (a, b) {
+    return a.toLowerCase().localeCompare(b.toLowerCase());
+  });
+  
+  //build alphabet hash
+  var alphabet = $('#alphabet').text();
+  var alphabetHash = {};
+  for (var j = 0; j < alphabet.length; j++){
+    var letter = alphabet[j];
+    alphabetHash[letter] = 0;
+  }
+
+  //see which letters are present in our terms
+  for (var k = 0; k < sortedTermsList.length; k++){
+    var firstLetter = sortedTermsList[k].charAt(0).toUpperCase();
+    alphabetHash[firstLetter] += 1;
+  } 
+
+  //clear page to then rebuild
+  $('.glossary-container').empty();
+
+  //add headers for each letter
+  for (var prop in alphabetHash) {
+    if (alphabetHash[prop] !== 0) {
+      var newIndexHeadline = '<h1>' + prop + '</h1>';
+      var newIndexContainer = $('<div class="index-container"></div>').attr('id', '' + prop).append(newIndexHeadline);
+      $('.glossary-container').append(newIndexContainer);
+    }
+  }
+
+  //add terms and definitions under appropriate letter headline
+  for (var l = 0; l < sortedTermsList.length; l++) {
+    var currentTerm = sortedTermsList[l];
+    var termHeadline = $('<h4>' + currentTerm + '</h4>');
+    var termSummary = terms[currentTerm];
+    //definition is a term as a headline and its corresponding paragraph summary
+    var newDefinition = $('<div class="term-container"></div>').append(termHeadline).append(termSummary);
+
+    //find definition's first character to find where to add it to the page
+    var nestUnder = sortedTermsList[l].charAt(0).toUpperCase();
+    $('#' + nestUnder).append(newDefinition);
+  }
+
+  //build table of contents (toc) then add to page
+  var toc = $('<div class="flex justify-content--space-between toc-container"></div>');
+  for (var letterInAlphabet in alphabetHash){
+    var newTOCletter = $('<span class="toc-index">' + letterInAlphabet + '</span>');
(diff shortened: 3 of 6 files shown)


KarenLoughrey:
@yesplease Thanks Bradley! Ideally I'd like to wait until I know copy is final; @srusonis do you know when you might have the list finalised? If finalisation is going to take a while and going live is high prio, I could start translations today and then update them once the source is finalised to speed the process up. It's not ideal, but it's also no problem. Let me know what you think!

yesplease:
@KarenLoughrey Thank you! Just got final approval from @srusonis on the copy so we should be good to go. cc @ihsekat

KarenLoughrey:
@yesplease  Thanks! Translations are in progress. Should be in tomorrow. Will let you know as soon as they're done :)

KarenLoughrey:
@yesplease Translations are done! :)

ihsekat:
@elliotykim Could take a look at this before you go on vacation? Pretty please? :sob:

yesplease:
@stewsmith can you review when you get back on Monday if Elliot can't get to it this weekend (which he shouldn't because weekend!)

website/optimization-glossary/index.hbs
```diff
@@ -0,0 +1,28 @@
+---
+layout: wrapper
+page_script: glossary.js
+layout_body_class:
+- glossary
+---
+<div id="main">
+  <div id="page-header">
+    <div class="container">
+      <h1>{{page_data.title}}</h1>
```
elliotykim: Noticed other pages use "visible_title":

https://github.com/optimizely/marketing-website/blob/master/boilerplates/page.hbs#L5

https://github.com/optimizely/marketing-website/blob/master/website/about/index.hbs#L4

website/optimization-glossary/index.hbs
```diff
@@ -0,0 +1,28 @@
...
+            <h4>{{this.TR_term}}</h4>
+            <p>{{this.TR_summary}} <a href="{{../linkPath}}{{this.link}}">  {{../page_data.cta}} ></a></p>
+          </div>
+        {{/each}}
+      </div>
+    </div><!--/.container-->
+  </div><!--/#content-->
+</div><!--/#main-->
+{{!-- Porting alphabet from YAML to sort elements on page regardless of language --}}
+<div id="alphabet" class="hidden">{{page_data.alphabet}}</div>
```
elliotykim: Another "container" for data like this is <input type="hidden" />:
http://www.w3schools.com/tags/att_input_type.asp

Don't feel compelled to change this now, but I noticed you're using .text() to read this value. If you have trouble with encoding for other languages, you might try a hidden input field and read the value with .val().

website/optimization-glossary/index.hbs
```diff
@@ -0,0 +1,28 @@
...
+      <h1>{{page_data.title}}</h1>
+    </div><!--/.container-->
+  </div><!--/#page-header-->
+  <div id="page-header-gradient"></div>
+  <div id="page-content">
+    <div class="container">
+      <div class="glossary-container">
+        {{#each page_data.glossary}}
+          <div class="term-container">
+            <h4>{{this.TR_term}}</h4>
```
elliotykim: Hmm. Once you merge with my branch, you will reference data in YML without the 'TR_' prefix. Let's plan on deploying this on Wed 8/19 with the rest of localization subfolders.

website-guts/assets/css/pages/glossary.scss
```diff
@@ -0,0 +1,21 @@
...
+  body.glossary {
+    .glossary-container {
+      max-width: 800px; 
+      margin: auto;
+    }
+    .term-container {
+      padding-bottom: 16px;
+    }
+  }
+}
```
elliotykim: Can you add an empty line at the end of the file? Just a stylistic nit

website-guts/assets/js/pages/glossary.js
```diff
@@ -0,0 +1,84 @@
+var alphabet = $('#alphabet').text();
+
+var buildTOCandSortedPage = function() {
```
elliotykim: Usually we would name this 'buildTocAndSortedPage' according to camel case convention.

Little more info about abbreviations:
http://stackoverflow.com/questions/15526107/acronyms-in-camelcase

website-guts/assets/js/pages/glossary.js
```diff
@@ -0,0 +1,84 @@
...
+  
+  //first gather all terms that are on page. terms may be in any of our languages.
+  var termsElements = $('.term-container');
+  
+  //build object with term and summarykey: "value",             
+  //key is string, value is the <p>  html element with summary of term
+  var terms = {};
+  for (var i = 0; i < termsElements.length; i++){
+    var term = $.text(termsElements[i].children[0]);
```
elliotykim: There's a strong assumption here that the first child element will always be the term. If the page structure changes in the future, that would require rewriting the JS. An easy solution is to give the h4 and p tags classes and reference them that way.

Also use jQuery to iterate the matched elements: https://api.jquery.com/each/

$('.term-container').each(function(index, element) {
  var term = element.find('.term').text();
  var summary = element.find('.summary').text();
  ...
});

elliotykim: Just for educational purposes, I wanted to show you how to create a dictionary of the terms in fewer steps.
```
var termsDict = {};
$('.term-container').each(function() {
  var term = $(this).find('.term').text();
  var summary = $(this).find('.summary').text();
  var firstLetter = term[0].toUpperCase();

  if (!(firstLetter in termsDict)) {
    // create new array
    termsDict[firstLetter] = [];
  }

  termsDict[firstLetter].push({
    term: term,
    summary: summary
  });
});
```

Then you could sort each array using a comparator like this:
```
for (var term in termsDict) {
  termsDict[term].sort(function(a, b) {
    return a.term.toLowerCase().localeCompare(b.term.toLowerCase());
  });
}
```

You can easily check if there are terms for a given letter. If the key exists in the dict, then there's a term!
```
var letters = alphabet.split(''); // convert string to array
letters.forEach(function(letter) {
  if (letter in termsDict) {
    // found! do something
  }
});
```

Anyways, hope that's helpful.

website-guts/assets/js/pages/glossary.js
```diff
@@ -0,0 +1,84 @@
...
+  //and sort them in a case-insensitive way
+  var sortedTermsList = Object.keys(terms).sort(function (a, b) {
+    return a.toLowerCase().localeCompare(b.toLowerCase());
+  });
+
+  //build alphabet hash
+  var alphabetHash = {};
+  for (var j = 0; j < alphabet.length; j++){
+    var letter = alphabet[j];
+    alphabetHash[letter] = 0;
```
elliotykim: This could be a boolean (true/false) since you're not using the count for anything. Makes assignment and the if statements below simpler.

`alphabetHash[firstLetter] += 1` --> `alphabetHash[firstLetter] = true`

`if (alphabetHash[letterInAlphabet] !== 0) { ... }` --> `if (alphabetHash[letterInAlphabet]) { ... }`

website/optimization-glossary/index.hbs
elliotykim: To be semantically correct, there's an HTML element intended for definition lists:
http://www.w3schools.com/tags/tag_dl.asp

website-guts/assets/css/styles.scss
```diff
@@ -33,6 +33,7 @@
 @import 'pages/events';
 @import 'pages/faq';
 @import 'pages/free-trial';
+@import 'pages/glossary';
```
stewsmith: Alphabetical order pls

website-guts/assets/js/pages/glossary.js
```diff
@@ -0,0 +1,84 @@
...
+  var terms = {};
+  for (var i = 0; i < termsElements.length; i++){
+    var term = $.text(termsElements[i].children[0]);
+    var summary = termsElements[i].children[1];
+    terms[term] = summary;
+  }
+
+  //now we need to sort terms. objects are by definition unsorted, so let's grab all terms in an array
+  //and sort them in a case-insensitive way
+  var sortedTermsList = Object.keys(terms).sort(function (a, b) {
```
stewsmith: Does sorting work as expected in other languages?

stewsmith:
@yesplease Nice job with this. I'll review more after you address @elliotykim 's comments.

website-guts/assets/css/pages/glossary.scss
stewsmith: What does the empty line do? My text editor automatically removes empty lines at the end of files. Judging from [your branch](https://github.com/optimizely/marketing-website/commit/8f3a8923dd08fafe8c435d61b8649d8fc0acf042?w=1#diff-293bc1b93834be2894f2e81e2e6bbbcfL9) maybe yours does too?

PR #1324, opened 2015-08-12
Fix typo from lead_scource to lead_source

yesplease:
Marketing Automation has been missing "lead_source___c" for a long time and we couldn't for the life of us figure out why, since it was always in our Segment debugging. Looks like we weren't seeing a discrepancy in our spelling. 

@stewsmith can you please review?

I think this typo catch should help.
@allisonsparrow once this is merged, can you keep your eye peeled going forward to see if this form field is now being populated in Marketo? I think this should do the trick.

This should address issues listed on https://optimizely.atlassian.net/browse/MKT-2009


stewsmith:
wow. lgtm.

:man: :hammer:

